PR #6, opened 2019-01-05
fix: add missing dependencies

tanguyantoine:
prettierFormat requires prettier to be present to format metadata output

Currently I get this error when using nexus. 
```
ERROR in ./node_modules/nexus/dist/prettierFormat.js
Module not found: Error: Can't resolve 'prettier' in 'PATH/node_modules/nexus/dist'
 @ ./node_modules/nexus/dist/prettierFormat.js 13:35-54
 @ ./node_modules/nexus/dist/metadata.js
 @ ./node_modules/nexus/dist/core.js
 @ ./node_modules/nexus/dist/index.js
 @ ./src/graphql/schema.ts
 @ ./src/graphql.ts
```

either we had prettier as a dependency or we ask developers to install it and add prettier to the `peerDependencies` section.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "url": "https://github.com/tgriesser"
   },
   "dependencies": {
+    "prettier": "^1.14.3",
     "tslib": "^1.9.3"
   },
   "devDependencies": {
@@ -42,7 +43,6 @@
     "husky": "^1.1.2",
     "jest": "^23.6.0",
     "lint-staged": "^7.3.0",
-    "prettier": "^1.14.3",
     "ts-jest": "^23.10.4",
     "ts-node": "^7.0.1",
     "tslint": "^5.11.0",


tgriesser:
I was actually hesitant on adding this formatting to the core. I'd like to avoid prettier in the `dependencies` or `peerDependencies`, since it's optional and only needed in development, and comes with a [massive](https://github.com/prettier/prettier/blob/5d515e5c61ad4875a590f69d6de576e02c38f2e7/package.json#L17-L74) long-tail of additional dependencies. 

The goal of nexus is to have zero dependencies and a single `peerDependency` of `graphql-js`. This is why it's dynamically `import`'ed - it's something the end user can add to their `devDependencies` if they want.

I wish there were a `peerDevDependencies` option. So anyway - thanks for the PR but would prefer this just be something added to documentation around using that feature. If you want to open a PR to add a more descriptive [error](https://github.com/graphql-nexus/nexus/blob/a77e6bb164f0609202339fe932fbae2c4ac61e07/src/prettierFormat.ts#L29), I think that'd be helpful! Otherwise I'll add some docs and a better descriptor explaining the rationale here.

tanguyantoine:
What about creating another package in that case ?

PR #9, opened 2019-01-05
build: add semantic-release

tanguyantoine:
add semantic release to facilitate package publishing


tgriesser:
Thanks, going to hold off on this until it gets closer to a 1.0 state (which should be soon). Still making some breaking changes in the api right now.

PR #11, opened 2019-01-21
Enum: allow picking with an alias when mixing

apalm:
I decided to take a stab at implementing this because it seems like a useful feature to have. Feel free to close if it isn't desirable :)

Please see https://github.com/apalm/nexus/blob/a8ee1a8763e6bc93a79db85931756d5ebabd86dd/src/__tests__/index.ts#L85 for usage.

### DIFF
--- a/src/__tests__/index.ts
+++ b/src/__tests__/index.ts
@@ -79,6 +79,26 @@ describe("nexus", () => {
       ).toEqual(["RED", "GREEN"]);
     });
 
+    it("can pick with mix and alias", () => {
+      const FavoriteColors = enumType<any>("FavoriteColors", (t) => {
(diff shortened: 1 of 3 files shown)


tgriesser:
Hmm, do you have an example of a case where you would find this useful? How would this be different than just defining the enum member directly?

apalm:
You're right; this probably isn't very useful for enums. My original intention was to implement this feature for objects, but I stopped because it looked like it mixing for objects is [not implemented yet](https://github.com/graphql-nexus/nexus/blob/develop/src/builder.ts#L335). I (think?) aliasing would be useful when mixing objects, since otherwise you'd have to redefine the output type of the field (please correct me if I'm wrong).

PR #12, opened 2019-01-21
[WIP] Implement mixing of objects

apalm:
If this was actually doable in 30 LOC, it probably would have been implemented already, so I'm probably missing something super obvious.

### DIFF
--- a/src/__tests__/index.ts
+++ b/src/__tests__/index.ts
@@ -139,18 +139,90 @@ describe("nexus", () => {
   });
 
   describe("objectType", () => {
-    const Account = objectType("Account", (t) => {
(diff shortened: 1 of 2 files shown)


tgriesser:
Thanks! Not something super obvious... mainly I'm still going back and forth on a few of the core API concepts, so type-mixing was one of those things I was going to tack on at the end. Should have everything nailed down around this by the beginning of next week, and then will be ready to accept PRs on things like this

tgriesser:
So actually decided to remove this feature for now, as referenced in #16 - can revisit if there are some good real-world use cases of situations where you can't accomplish the same sort of thing with JS.

PR #31, opened 2019-02-12
Update library-authors.md

thebergamo:
From my understanding the first example in this pages was using a variable that doesn't exist in any context and could let to few headaches when someone just copy/paste the example code.

So I just make it using the proper variable.

It might be wrong assumption from my side, please let me know ;)


tgriesser:
Great, thanks @thebergamo!

thebergamo:
Glad to help :)

PR #32, opened 2019-02-12
Authorization Field

tgriesser:
Working off this as a separate ticket from #23, going to publish this as a `0.10.0-alpha.1` just so I can play around with it in real-world use and get a feel for it.


tgriesser:
Merging this as a first step of a more robust authorization layer in Nexus, will release under 0.10.0

PR #47, opened 2019-02-20
Fix docs for shouldGenerateArtifacts default

ChristianBoehlke:
Hello,

Thank you for the great work with Nexus!

I have noticed that the docs state the opposite default value for `shouldGenerateArtifacts` as it is defined in the code:
https://github.com/prisma/nexus/blob/a6156ee8f46b2e525c24b849934a51e315f4ece1/src/builder.ts#L1012-L1014


tgriesser:
Great, thanks @ChristianBoehlke!

PR #49, opened 2019-02-21
fix missing comma in code example

nikiask:


nikiask:
Also, where can I import `Kind` from?

tgriesser:
Thanks!

> Also, where can I import `Kind` from?

What is `Kind`?

nikiask:
In the example there's a `Kind.INT`
```typescript
const DateScalar = scalarType({
  name: "Date",
  asNexusMethod: 'date',
  description: "Date custom scalar type",
  parseValue(value) {
    return new Date(value);
  },
  serialize(value) {
    return value.getTime();
  },
  parseLiteral(ast) {
    if (ast.kind === Kind.INT) {
      return new Date(ast.value);
    }
    return null;
  },
});
```

ChristianBoehlke:
> Also, where can I import `Kind` from?

[`Kind`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/094be5661ef104a6533b6521fb639db7c36b214c/types/graphql/language/kinds.d.ts) comes from `@types/graphql`. All three methods `parseValue`, `serialize`, and `parseLiteral` are defined there as well on the [`GraphQLScalarTypeConfig`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/094be5661ef104a6533b6521fb639db7c36b214c/types/graphql/type/definition.d.ts#L295-L306) interface.
You can import it with `import { Kind } from 'graphql';`.

Issue #51, opened 2019-02-21
extend Input


israelglar:
It's possible to extend any Type that comes from prisma with `extendType`. That's a great feature of nexus! But it's not possible to do something like `extendInputType`.

If I use `extendType` on a InputType like this:
```ts
const createAccountInput = extendType({
  type: 'AccountCreateInput',
  definition(t): void {
    t.string('password', {nullable: false});
  },
});
```
I get this error:
```ts
Expected AccountCreateInput to be a possible input type, saw GraphQLObjectType
```

#### My Question/Feature request
It there a way to extend Input Type?

tgriesser:
This is something I should be able to whip up pretty quickly. I'm curious what the cases are where you'd want to do this though rather than just defining on the type to begin with?

israelglar:
This came up because of the architecture of my app. 
My graphql schema depends on the user input. 

For example: If the input it's just 'Account', then a schema will be created like this:
```gql
type Account {
  name: string 
}
input AccountCreateInput {
  name: String
}
```

But if the input is ['Account', 'Password'], then the schema will be:
```gql
type Account {
  name: string 
  password: string 
}
input AccountCreateInput {
  name: String
  password: string 
}
```

This password example it's not the best, because it doesn't make much sense x.x but that's the idea :p

I found that nexus is the best library for what we are doing, because we already use prisma as an ORM. We plan to use `extendType` a lot! The only missing part is to extend inputs.

I can contribute with a PR with some help :)

israelglar:
But I also think that feature makes sense to have in the library because maybe you want a field to show on a mutation arg in your graphql, but don't want that field to be persisted in the database.

We can already do that (resolved fields) to results of queries and mutations, but it's a similar logic to the arguments

tgriesser:
See how https://github.com/prisma/nexus/pull/54 looks

israelglar:
It looks great :D It really is a simple feature, but really useful!
When will it be updated on npm?

tgriesser:
Released in 0.9.16

blazestudios23:
Hey can you do some documentation on how to include the old fields and add new ones when doing a extendInputType? When I try this it just overwrites the input type with new fields.

PR #54, opened 2019-02-22
Add extendInputType

tgriesser:
#51


blazestudios23:
Hey can you do some documentation on how to include the old fields and add new ones when doing a extendInputType? When I try this it just overwrites the input type with new fields.

PR #61, opened 2019-02-27
create folder path recursively if missing when generating files

kandros:
before this, if we specified a path to writing generated files that don't exist the file creation fails.

```
{ [Error: ENOENT: no such file or directory, open '[....removed]/src/graphql/schema/local/__generated__/schema.graphql']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path:
   '[....removed]/src/graphql/schema/local/__generated__/schema.graphql' }
```


tgriesser:
Thanks! I believe the `recursive` option for was only added recently, but I presume this shouldn't cause any errors if running on older node versions.

PR #63, opened 2019-03-01
fix: generate schema and types only if not in production

Knaackee:
The documentation states that the schema and types gets generated if not in production environment, but the default in the source code was `process.env.NODE_ENV === "development"`.

I encountered this because i was using it with `process.env.NODE_ENV === "test"` in testing mode and no schema and types was generated.

### DIFF
--- a/docs/api-makeSchema.md
+++ b/docs/api-makeSchema.md
@@ -41,7 +41,7 @@ interface SchemaConfig {
     | false;
   /**
    * Whether the schema & types are generated when the server
-   * starts. Default is !process.env.NODE_ENV || process.env.NODE_ENV === "development"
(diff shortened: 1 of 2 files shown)


tgriesser:
Thanks for the PR @Knaackee - you can customize this behavior for your schema by setting `shouldGenerateArtifacts` on the `makeSchema`:

```ts
const schema = makeSchema({
   shouldGenerateArtifacts: process.env.NODE_ENV !== 'production',
   // other config ...
})
```

I'd rather not include other environments than development, because IMO the default behavior should be an error if the types aren't up to date when testing, in the same way that snapshot tests shouldn't be updated on a CI run. I'll make an update in the docs to clarify.

PR #65, opened 2019-03-03
Prevent reordering generated imports by deterministically sorting by alias name

jacobwgillespie:
Since the typegen import paths are resolved asynchronously, they add their config to `importsMap` in a non-deterministic order.  This means that when `Object.keys(importsMap)` is called, it's not guaranteed that the imports are returned in the same order, which in turn causes the generated file to sometimes re-shuffle imports.  This hampers the ability to check the generated files into version control, since the imports are occasionally re-ordered.

### Reproduction

1. Have more than one file in `typegenAutoConfig.sources`:

    ```typescript
    makeSchema({
      ...
      typegenAutoConfig: {
        // at least two source files here
        sources: [
          {
            alias: 'one',
            source: 'path/to/one',
          },
          {
            alias: 'two',
            source: 'path/to/two',
          },
          ...
        ],
      },
    })
    ```

2. Run the app multiple times - the generated typegen file will sometimes import `one` first and sometimes `two` will be imported first:

    ```typescript
    // sometimes
    import * as one from 'path/to/one'
    import * as two from 'path/to/two'

    // sometimes
    import * as two from 'path/to/two'
    import * as one from 'path/to/one'
    ```

### Fix

This PR sorts the keys of `importsMap` (the import aliases) so that the ordering will remain deterministic across multiple runs.

### DIFF
--- a/src/typegenAutoConfig.ts
+++ b/src/typegenAutoConfig.ts
@@ -319,7 +319,7 @@ export function typegenAutoConfig(options: TypegenAutoConfigOptions) {
 
     const imports: string[] = [];
 
-    Object.keys(importsMap).forEach((alias) => {
+    Object.keys(importsMap).sort().forEach((alias) => {
       const [importPath, glob] = importsMap[alias];
       imports.push(
         `import ${glob ? "* as " : ""}${alias} from "${importPath}"`


tgriesser:
Ah, good catch! Thanks for the detailed explanation and PR!

jacobwgillespie:
No problem, thanks for the fast merge!

Issue #21, opened 2019-02-08
Subscription Example/Support?


williamluke4:
Are subscriptions currently supported and if so are there any examples?

ChristophDietrich:
I'm also very interested in the subscription examples. Awesome work so far!

tgriesser:
Subscriptions actually [still need to be implemented](https://github.com/prisma/nexus/blob/develop/src/definitions/objectType.ts#L118-L122).

Most of it should be as simple as following the patterns in `objectType` and adding a separate `subscribe` field & typings for that. I'll probably get around to it in the next week or two, but if anyone wants to dig into the library a bit and take a shot at it though, a PR would be welcome!

BjoernRave:
How is the status of this? I am really looking forward to use Subscriptions with nexus and nexus-prisma

kandros:
@tgriesser is there anything we can help you with?

tgriesser:
Started doing a little investigation into this. Can folks who use subscriptions clarify what their expectations might be for an API here? I've never used GraphQL subscriptions so I want to make sure I'm understanding the use case/how the types fit together. I think it might need to be a bit different than the normal patterns.

#### Proposed API:

```ts
export const newSubField = subscriptionField({
  type: SomeType,
  subscribe() {
    // ...
  },
  resolve() {
    // ...
  }
})
```

#### Open Questions:

1. What is the signature of subscribe?

```ts
subscribe(root, args, ctx, info): AsyncIterator<T>
```

What is the type of the `root` here? Is it empty object, similar to `Query`/`Mutation`? Is `root` ever even useful here, or should we do something similar to https://github.com/taion/graphql-relay-subscription and just omit it: `subscribe(args, ctx, info)`? 

Is `ctx` always guaranteed to be the same type as the one in normal query resolvers, or can/should it be different? 

What is `T` in the above return, and where should it come from? Should the developer provide this as it works with other typings or can it be inferred?

2. What is the signature of resolve?

```ts
resolve(root, args, ctx, info)
```

What is `root` here, is this the type T from the above subscribe?

Is it reasonable that we require a `resolve` always be provided alongside the `subscribe`, even if it's just an identity function `resolve: root => root`

I don't think this should take too much to add, just wanted to be clear on the api we're looking for here.

tgriesser:
Also, another question: is the only time you'd ever use/implement `subscribe` on the root `Subscription` type?

tgriesser:
Just opened #66

[...]

PR #66, opened 2019-03-03
Add subscriptionField, initial pass at #21

tgriesser:
Interested to get feedback on the implementation here, it works slightly differently than the other types  since subscription has different semantics:

TODO:
- [ ] Add subscription example to kitchen sink

### DIFF
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -13,6 +13,7 @@ import {
   NexusWrappedType,
 } from "./wrapping";
 import { BaseScalars } from "./_types";
+import { GraphQLFieldResolver } from "graphql";
 
 export interface CommonFieldConfig {
   /**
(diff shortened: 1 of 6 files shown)


src/definitions/definitionBlocks.ts
```diff
@@ -76,6 +83,7 @@ export interface NexusOutputFieldConfig<
 
 export type NexusOutputFieldDef = NexusOutputFieldConfig<string, any> & {
   name: string;
+  subscribe?: GraphQLFieldResolver<any, any>;
```
Devlin556: Why the `subscribe` field should have `GraphQLFieldResolver` type here?

nevescode:
I'm really looking forward to it! 👍

Knaackee:
When did subscription support will get released? I really need this.
Thanks for your great work!

tgriesser:
Released this along with a few other changes in 0.11

I'm calling the `mutationField` feature alpha. It should work but it's undocumented for now. Kick the tires on it and let me know what you think!

PR #67, opened 2019-03-04
Update playground url

estrada9166:
+ Update playground url to prevent that if you're on `docs` and click on `playground` the url is going to be `https://nexus.js.org/docs/playground`

+ Replace `./nexus` for `nexus` to follow the same order on all the docs.


tgriesser:
Nice, thanks @estrada9166!

PR #73, opened 2019-03-08
core: expose subscriptionField & add documentation

Gomah:
- expose `subscriptionField`
- add `subscriptionField` documentation

Related issues / pull requests:
- [x] #21 
- [x] #66


tgriesser:
Hey, thanks for the PR @Gomah - just pushed 0.11.1 with the `subscriptionField` publicly accessible. 

Didn't merge this PR since the `extendType` actually won't work the same way for `Subscription` as it does for other types, as other fields don't have the `subscribe` property available.

Also, the docs need to be a little different for `subscriptionField` then they are for `mutationField` / `queryField`, since the signature is a little different and it's not just a passthrough to `extendType`.

If you want to open another PR with more detailed documentation about the expectations/signature around subscriptions, etc. that'd be great!

gotexis:
+1 for this

PR #76, opened 2019-03-08
fix mkdir throwing on node 10.9 below

### LINKED ISSUE #75: makePrismaSchema does not generate code due to an exception from mkdir
I'm trying to run the `prisma-examples/typescript/graphql/` example.

I think the problem is in the following code:
https://github.com/prisma/nexus/blob/7dc2ac7aedc7992fb709ed580a13435993aeffe7/src/typegenMetadata.ts#L93

which causes:
```
{ Error: EEXIST: file already exists, mkdir '/home/test/prisma-examples/typescript/graphql/src/generated'
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/home/test/prisma-examples/typescript/graphql/src/generated' }
```

kandros:
fixes #75

### DIFF
--- a/src/typegenMetadata.ts
+++ b/src/typegenMetadata.ts
@@ -69,10 +69,11 @@ export class TypegenMetadata {
     }
     const fs = require("fs") as typeof import("fs");
     const util = require("util") as typeof import("util");
-    const [readFile, writeFile, mkdir] = [
+    const [readFile, writeFile, mkdir, exists] = [
       util.promisify(fs.readFile),
       util.promisify(fs.writeFile),
       util.promisify(fs.mkdir),
+      util.promisify(fs.exists),
     ];
     let formatTypegen: TypegenFormatFn | null = null;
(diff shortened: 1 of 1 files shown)


kandros:
updated with recommendations from node fs docs

src/typegenMetadata.ts
```diff
@@ -89,8 +89,17 @@ export class TypegenMetadata {
       readFile(filePath, "utf8").catch(() => ""),
     ]);
     if (toSave !== existing) {
-      const dirPath = path.dirname(filePath)
-      await mkdir(dirPath, { recursive: true })
+      const dirPath = path.dirname(filePath);
+      try {
+        await mkdir(dirPath, { recursive: true });
+      } catch (e) {
+        if (e.code === "EEXIST") {
```
Weakky: `if (e.code !== 'EEXIST') { throw e; }` seems better here

PR #80, opened 2019-03-11
Fix apollo-fullstack example

TheHalcyonSavant:
Without these changes apollo-fullstack example will fail when you run 'npm start'


tgriesser:
Thanks @TheHalcyonSavant - would you be able to add a .gitignore / remove the `store.sqlite` file?

TheHalcyonSavant:
@tgriesser can you clarify what is your intent?

1. store.sqlite is copied here from https://github.com/apollographql/fullstack-tutorial/tree/master/final/server. How would the example work if I remove store.sqlite? Is there any other way to fix if I remove store.sqlite from my PR?

2. Why to add to .gitignore and where? Do u want me to add a new file: examples/apollo-fullstack/.gitignore?

Sorry, but I need to understand first why before I do any changes.

tgriesser:
Oh my mistake, I assumed it was auto-generated. Thanks for the fix!

PR #81, opened 2019-03-11
Extract reusable type for the "type" option

agonbina:
This allows you to pass a `NexusWrappedType` instance as an argument for type when using `arg({ type: ... })` utility to define an argument for a type.

Ex: 

```
import MyType from './MyType'

const MyOtherType = queryField('name', {
  args: { myType: arg({ type: MyType }) } 
})
```

The feature actually works as is, the type is incorrect though.


tgriesser:
Looks great! Thanks for finding and fixing this!

agonbina:
🎉

PR #84, opened 2019-03-14
Fix import path generation on Windows

### LINKED ISSUE #41: typeGenAutoConfig wrong relative path in Windows
These lines generate the imports of the typegenAutoConfig. https://github.com/prisma/nexus/blob/18a306ee2903d98fc110c1067e1903ea0ee5f256/src/typegenAutoConfig.ts#L212-L215  And use this function https://github.com/prisma/nexus/blob/18a306ee2903d98fc110c1067e1903ea0ee5f256/src/typegenAutoConfig.ts#L340-L350

The problem is with this line:
```ts
    return `.${path.sep}${path.join(relative, filename)}`;
```

In Windows it generates an import statement with backslashes `\`, and when node tries to compile it, it can't.

Maybe change it to:
```ts
    return `./${path.join(relative, filename)}`;
```

ghalle:
The `typeGenAutoConfig` currently generates invalid import paths containing non-escaped backslashes if run on Windows.

This PR fixes that by using a regex that replaces every backslash with a forward slash.

Fix #41

### DIFF
--- a/src/typegenAutoConfig.ts
+++ b/src/typegenAutoConfig.ts
@@ -321,8 +321,9 @@ export function typegenAutoConfig(options: TypegenAutoConfigOptions) {
 
     Object.keys(importsMap).sort().forEach((alias) => {
       const [importPath, glob] = importsMap[alias];
+      const safeImportPath = importPath.replace(/\\+/g, '/');
       imports.push(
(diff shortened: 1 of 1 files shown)


tgriesser:
Thanks!

PR #105, opened 2019-04-08
Fix library authors example's syntax issue

leonardodino:
There were some syntax and scope issues on the example.
This should fix it.

(the code wouldn't without this fix, there were mismatched braces)

It would be nice to run prettier on the markdown, it can catch some of this bugs (: and provides a more consistent style throughout the docs

### DIFF
--- a/docs/library-authors.md
+++ b/docs/library-authors.md
@@ -17,33 +17,30 @@ export const UserConnectionTypes = connectionType("User");
 Where `connectionType` is really just a wrapper creating two `objectTypes`:
 
 ```ts
-import { core } from 'nexus';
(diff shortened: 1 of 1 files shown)


tgriesser:
Thanks for the fixes! 

> It would be nice to run prettier on the markdown, it can catch some of this bugs (: and provides a more consistent style throughout the docs

Hmm, I thought it should have been running - I have 

```
"lint-staged": {
    "*.{js,json,css,md}": [
      "prettier --write",
      "git add"
    ]
  },
```

in the `package.json`, so I'm not sure why it wouldn't be running? Though looking at that config again now that should definitely also include `.ts` 🤦‍♂️

Issue #70, opened 2019-03-05
How to do schema stitching with Nexus?


otrebu:
Hi there!

This is a question not an issue or a bug 😄 

What is it the best way to do schema stitching with nexus?
The only thing I can think about is using the converter to convert the schemas to code first, but it doesn’t sound ideal and I am not sure if it is possible to use fragment with it.

Thanks for the amazing work!

pantharshit00:
I think we can just use graphql-tools for this.

Since the output of nexus is a graphql js schema one can easily use packages from the existing ecosystem.

```js
import {
  makeExecutableSchema,
  addMockFunctionsToSchema,
  mergeSchemas,
} from 'graphql-tools';
import { makeSchema } from 'nexus';


const nexusSchema = makeSchema({
  types: [Account, Node, Query, StatusEnum],
  // or types: { Account, Node, Query }
  // or types: [Account, [Node], { Query }]
});

const authorSchema = makeExecutableSchema({
  typeDefs: `
    type User {
      id: ID!
      email: String
    }

    type Query {
      userById(id: ID!): User
    }
  `
});

addMockFunctionsToSchema({ schema: authorSchema });

export const schema = mergeSchemas({
  schemas: [
    chirpSchema,
    nexusSchema
  ],
});

```

otrebu:
@pantharshit00 oh that is great thank you!

Just to clarify a couple of things:

`addMockFunctionsToSchema({ schema: authorSchema });` this is optional and just for mocking/testing right?

```
export const schema = mergeSchemas({
  schemas: [
    chirpSchema,
    nexusSchema
  ],
});
```
Did you mean `authorSchema`  rather than `chirpSchema`?

😄

tgriesser:
@otrebu there are some good solutions for doing schema extension and composition with other GraphQL schemas in Nexus that I need to document / add a few new options for. 

Unfortunately don't have time to fully go into them right this minute, but I'll leave this ticket open and report back once I have some time to document/add some examples for how to do this.

Would you be able to share the general structure of how your schema is split up and how you'd expect a tool that allowed schema stitching with nexus to work? Would most types be written in nexus and only some would be stitched in? 

Also, have you seen the [extendType](https://nexus.js.org/docs/api-extendtype) api for building a schema from multiple places in the codebase? There's also a [mutationField](https://nexus.js.org/docs/api-mutationfield) and I'll soon add a similar `queryField` api for this use case.

kandros:
[...]

Issue #88, opened 2019-03-15
Interop with graphql-js


janicduplessis:
I'm investigating using this to progressively replace `graphql` to generate the schema of a large application. Using types defined with `graphql` works well by referencing them using their name but I haven't found I way to use types defined with `nexus` inside a type defined with `graphql`.

I did try using `extendType` to extend my types defined using `graphql` but that doesn't seem to work currently. I'm getting a duplicate type error: `Error: Schema must contain unique named types but contains multiple types named "Viewer".`.

Here's a simplified example
```js

const Viewer = new GraphQLObjectType({
  name: 'Viewer',
  fields: {
    name: { type: GraphQLString },
  },
});

// Using extendType to extend types defined with `graphql`
const NexusViewer = extendType({
  type: 'Viewer',
  definition(t) {
    t.int('age');
  },
});

const Query = new GraphQLObjectType({
  name: 'Query',
  fields: () => ({
    viewer: { type: Viewer },
  }),
});

const schema = makeSchema({
  types: [
    Query,
    NexusViewer,
  ],
});
``` 

Or alternatively provide a way to convert to a `graphql` type.

```js
const NexusObjectType = objectType({ name: 'NexusObject' definition() { /* ... */ } });

const GraphQLObjectType = new GraphQLObject({
  name: 'GraphQLObject',
  fields: { nexus: { type: toGraphQLType(NexusObjectType) } }
});
```

tgriesser:
Yep, this ties into #70 - there needs to be a better story around interop between the two - I'm going to be working more on this in the next few weeks, will comment here once I have something more well defined to test out fo this.

stalniy:
I created an example which shows how `graphql` types can be used together with nexus: https://codesandbox.io/s/zqq7p9w1jp

In order to do this:
1. I extended nexus `SchemaBuilder` and overloadded `getOrBuildType`
    
    ```js
    getOrBuildType(name, fromObject = false) {
       if (isType(name)) {
         return this.finalize(name);
       }

       return super.getOrBuildType(name, fromObject);
     }
    ```
2. And later you can easily reference native graphql types during nexus schema building process

tgriesser:
Check out #143 and see if this works any better with the newly published `0.12.0-beta.1`

thehappycoder:
@tgriesser If you still remember, can you point to the line in that huge PR?

blazestudios23:
[...]

PR #110, opened 2019-04-10
Parse graphql-js types into Nexus structures

tgriesser:
Still needs tests to confirm everything is working correctly, but this should allow better interop between existing graphql-js types and nexus by first parsing to nexus compatible structures before processing, so things like type-extension will work correctly.

Fixes: #70, #88

/cc @janicduplessis @otrebu @Weakky


stalniy:
Could you please add some explanation because it looks strange for me?

So, let me explain how I understood the changes:
You have built graphql-js scheme which you convert into nexus SchemeBuilder blocks in order to convert this again into graphql-js scheme. Am I right?

tgriesser:
Yeah, thought about this a little more and agree it is confusing. Closing in favor of #143 which will solve the same underlying issue.

PR #120, opened 2019-04-20
Add promise support to subscriptions

### LINKED ISSUE #119: Cannot return complete subscription payload
It's only possible to return the `node` in the `subscribe` function right now but we need to return the entire payload with the `mutation` which is part of Prismas `SubscriptionPayload`.
I think the problem is that the subscribe function is not able to resolve a promise which can be returned.

I'm getting the following error
```typescript
Type '(parent: object, { userId }: any, ctx: Context) => UserSubscriptionPayloadSubscription' is not assignable to type '(root: object, args: any, ctx: Context, info: GraphQLResolveInfo) => AsyncIterator<any>'.
        Property 'next' is missing in type 'UserSubscriptionPayloadSubscription' but required in type 'AsyncIterator<any>'.
```

```typescript
export interface UserSubscriptionPayloadSubscription
  extends Promise<AsyncIterator<UserSubscriptionPayload>>,
    Fragmentable {
  mutation: () => Promise<AsyncIterator<MutationType>>;
  node: <T = UserSubscription>() => T;
  updatedFields: () => Promise<AsyncIterator<String[]>>;
  previousValues: <T = UserPreviousValuesSubscription>() => T;
}
```

```typescript
export const userSubscription = subscriptionField("user", {
  type: UserSubscriptionPayload,
  args: {
    userId: idArg(),
  },
  subscribe(parent, { userId }, ctx) {
    return ctx.db.$subscribe.user({
      node: {
        id: userId,
      },
    });
  },
  resolve: (payload) => payload,
});
```

sapkra:
With this PR I'm adding the ability to return a promise in the `subscribe` function and should make it possible to return the entire `SubscriptionPayload` of Prisma.

I can also provide an example if you want...

Fixes #119

### DIFF
--- a/src/definitions/subscriptionField.ts
+++ b/src/definitions/subscriptionField.ts
@@ -26,7 +26,7 @@ export interface SubscribeFieldConfig<
     args: ArgsValue<TypeName, FieldName>,
     ctx: GetGen<"context">,
     info: GraphQLResolveInfo
-  ): AsyncIterator<T>;
+  ): MaybePromise<AsyncIterator<T>> | MaybePromiseDeep<AsyncIterator<T>>;
 
   /**
    * Resolve method for the field


andres-valdes:
Absolutely in agreement here, just ran into this issue today. If using subscriptions for anything non-trivial and asynchronous this problem arises extremely quickly.

It looks like this was added to graphql-js in v0.11.0.
https://github.com/graphql/graphql-js/releases/tag/v0.11.0

src/definitions/subscriptionField.ts
```diff
@@ -26,7 +26,7 @@ export interface SubscribeFieldConfig<
     args: ArgsValue<TypeName, FieldName>,
     ctx: GetGen<"context">,
     info: GraphQLResolveInfo
-  ): AsyncIterator<T>;
+  ): AsyncIterator<T> | Promise<AsyncIterator<T>>;
```
andres-valdes: Should we use
```suggestion
  ): MaybePromise<AsyncIterator<T>> | MaybePromiseDeep<AsyncIterator<T>>;
```
here like we do for FieldResolver?

tgriesser:
Is this an actual runtime issue or just a types issue? The code changed here actually will define the field asynchronously, which isn't possible and is incorrect. We don't have any tests around subscriptions, which is the reason this passed CI.

I believe just changing the type signature of `subscribe` from `AsyncIterator<T>` to `MaybePromise<AsyncIterator<T>> | MaybePromiseDeep<AsyncIterator<T>>` is all you need to do. As @andres-valdes pointed out, this is the behavior in GraphQL, so it should take care of the rest.

src/definitions/subscriptionField.ts
sapkra: Seems to work fine.

Issue #118, opened 2019-04-19
Allow TypeScript enums to be linked with AutoConfig


Nayni:
I'm currently struggling with using and `enum` defined in TypeScript which is used as members for an `enumType` in nexus.

The problem is as follows:

I have an existing TypeScript enum: 
```typescript
export enum Category {
    CATEGORY_A = "CATEGORY_A ",
    CATEGORY_B = "CATEGORY_B ",
    CATEGORY_C = "CATEGORY_C ",
    CATEGORY_D = "CATEGORY_D "
}
```

The values of this enum will also be exposed in the schema and thus I created the following `enumType`:
```typescript
export const CategoryEnum = enumType({
    name: "Category",
    members: [
        Category .CATEGORY_A ,
        Category .CATEGORY_B ,
        Category .CATEGORY_C ,
        Category .CATEGORY_D 
    ]
});
```

This works fine, and nexus will understand that my enums are strings and will generate the schema correctly. It will also generate the type definitions for this and they will be:
```typescript
export interface NexusGenEnums {
  Category: "CATEGORY_A " | "CATEGORY_B " | "CATEGORY_C " | "CATEGORY_D "
}
```

And this is where the problem comes into play. Because now when I setup a resolver the incoming type of for example an input will be mapped to the nexus generated type which is a string union type. However in our implementation we always refer to our own TypeScript enum and the generated nexus type doesn't fit into this type because strings do not magically typecheck into an enum (the other way around works fine but not when the input is a string).

I also tried to use the `typegenAutoConfig` and point nexus to our existing enum but it doesn't seem to pick up the alias and keeps using the string union. (even if I change the Regex to include looking for enum)

Is this an expected behavior and if so, would it maybe be a useful feature to also allow TypeScript enums to be configurable as candidates for type alias?

PS: Love the library btw, been working on migrating our codebase and it's been super enjoyable so far.

Nayni:
Did some fiddling with my own fork and came up with the following to allow enums to have backing types (only tested in isolation so far):
https://github.com/prisma/nexus/compare/develop...Nayni:allow-enum-backing-types

Is this something that could be considered @tgriesser ? I'm not sure why enums were excluded from having backing types in the first place?

tgriesser:
Hmm, yeah I don't know why it wasn't considered - probably just an oversight. Feel free to open a PR and I'll review it a little closer. I think we might want this to exclude `const enum`s if possiblee?

Nayni:
Solved by #142 
[...]

PR #142, opened 2019-05-17
Allow enum backing types

Nayni:
This should allow for enums to be mapped as backing types (see #118). In order to keep things clear I separated the regex's for enums from types/interfaces/classes. I also added a new set of tests to verify the implementations.

As suggested in #118 I excluded `const enum`s but would still question why these should be excluded since backingtypes are purely for typing, I can't think of a reason why not (I barely use `const enum`s myself so I might be missing something)?


tgriesser:
> As suggested in #118 I excluded const enums but would still question why these should be excluded since backingtypes are purely for typing, I can't think of a reason why not (I barely use const enums myself so I might be missing something)?

Yeah maybe these don't need to be excluded. Feel free to take that out

tgriesser:
I'm also wondering whether you could include in the PR a clean way to make this possible:

```ts
export enum A {
  ONE = 1,
  TWO = 2,
}

const EnumA = enumType({
  name: "A",
  members: A,
})
```

I realize this will likely work out of the box for your example with `ONE = "ONE"...` because the name and value are both strings, but when they're not TypeScript turns the above into:

```ts
var A;
(function (A) {
    A[A["ONE"] = 1] = "ONE";
    A[A["TWO"] = 2] = "TWO";
})(A || (A = {}));
```

Since numbers aren't valid enum names it'll fail. Could it be as simple as filtering the keys to remove invalid enum names when constructing, and this would maintain the correct mapping?

Nayni:
If we consider your example @tgriesser : 
```typescript
export enum A {
  ONE = 1,
  TWO = 2,
}

const EnumA = enumType({
  name: "A",
  members: A,
}
```
What would be the preferred outcome? Since `nexus` is mainly a type layer I would imagine that the only thing we can do is try to create a valid `GraphQLEnumType` from the TypeScript enum. This would means we have 2 options as far as I can see:
* Make this fail and only allow a TypeScript enum (or a single value) if the value(s) are all strings,
* Parse the TypeScript enum value (to string, so in this case `1` becomes `"1"`) and use it as a key for the `GraphQLEnumType` which will use it as a string BUT this might cause confusion and problems as I would also expect the `GraphQLEnum` value be a string then was used a input?

tgriesser:
> I missed the existing functionality of using internal mapped values. In this case I would think that just mapping over the enum and creating a map similar to

Yep! Exactly what I was thinking.

src/definitions/enumType.ts
```diff
@@ -1,6 +1,10 @@
 import { NexusTypes, withNexusSymbol } from "./_types";
 import { assertValidName } from "graphql";
 
+type TypeScriptEnumLike = {
```
Nayni: I'm not 100% happy with this but I couldn't find a better way to allow a TypeScript enum definition to fit into the typing, open to better suggestions

Nayni:
Now that it's possible to pass a TS enum into the members property, should we still allow `const enum`s as backing types? It might send a mixed message because `const enum` is not going to fit into the members config like a regular `enum`. 

On the other hand it could still be useful to make the typings match (even if you have to manually write out all the members)

tgriesser:
> Now that it's possible to pass a TS enum into the members property, should we still allow const enums as backing types? It might send a mixed message because const enum is not going to fit into the members config like a regular enum.

Yeah that's why I had initially said that about not allowing them, though I don't think it'll necessarily be a problem, since folks using `const enum` will likely know TypeScript well enough to understand the limitations.

src/definitions/enumType.ts
tgriesser: This is the best solution I can find as well, so I think it's probably alright.

src/builder.ts
```diff
@@ -490,11 +490,13 @@ export class SchemaBuilder {
         }
       });
     } else {
-      Object.keys(members).forEach((key) => {
-        values[key] = {
-          value: members[key],
-        };
-      });
+      Object.keys(members)
+        .filter((key) => isNaN(+key)) // Ensures we only consider enum keys in case a TypeScript enum was passed as the members property.
```
tgriesser: Symbols can technically be keys of a mapping, though probably rare we may want to guard against that case.

tgriesser: We may also want to check not just if it's not a string, but also that `members[key] === members[members[key]]` which would indicate that it's actually a typescript enum, rather than a user-defined key in a normal object that happens to be an invalid value. 

Along those lines, I think it'd be worthwhile to add GraphQL's `assertValidName` guard on the `key` of the mapping similar to how we do it elsewhere for the object & field names.

Nayni: > Symbols can technically be keys of a mapping, though probably rare we may want to guard against that case.

Do you mean it's possible that Symbols are the keys of `Object.keys` and that we should guard against having them as member keys as input for `GraphQLEnumType`?

tgriesser: Actually, I thought Symbols could be values of an enum but they can't, so you can ignore that.

Nayni: I added an `assertValidName` check to ensure the enum name is valid in the end. The suggest members check is not really feasible because that check would only work for enums that have a numeric value, and since enum names of a TS enum are always strings the `isNaN` check is actually the more reliable and simple way I found to ensure we grab the right keys.

I also improve the test assertions to ensure both names and values are what is to be expected.

tgriesser:
Looks like there's a minor issue due to the changes merged in #143 - if you want to take a look at getting it sorted, feel free. Otherwise I will get it merged tomorrow and get this out with an 0.12.0 release.

Nayni:
I think this should do it!

PR #143, opened 2019-05-18
Initial pass at dynamic field inclusion/definition

### LINKED ISSUE #22: Relay style connections example
Very interesting library, thanks to everyone involved!

I'd be interested in a relay-style connection example! How'd you do these?

### LINKED ISSUE #136: Type auto registration during scheme build
I’m building dynamic GraphQL scheme. 

The issue is that in some cases I can’t know the type before hand, so can't pass it into `types` option of `makeSchema` function but I can build it lazily during scheme building. However there is no way to register a new type during scheme build process, also nexus doesn’t register unknown types found on `field.type` as `graphql-js` does.

**Expected behavior**:
nexus should register types found in `field.type` automatically (similar to graphql-js)

**Nexus example**:

It's very simplified just to show the issue
https://codesandbox.io/s/rjloyj3q4n

It throws 
```
/*
 * Error: Looks like you forgot to import MyQuery
 *        in the root "types" passed to Nexus makeSchema
 */
```

**graphql-js Example**

The same example works correctly if using `graphql-js`: https://codesandbox.io/s/ojpv91ojl6

### LINKED ISSUE #137: Type 'NexusInputObjectTypeDef<string>' is not assignable to type 'NexustArgDef<string>'.
Can't find anything in docs about how to create and register input objects.

What i want to achieve is schema like this:

`type Query{
signup(input: SignupInput): AuthPayload
}`

when i do:

<pre>export const Mutation = mutationType({
  definition(t) {
    t.field('signup', {
      type: 'AuthPayload',
      args: {
        input: inputObjectType({
          name: 'RegisterInput',
          definition: (t)=>{
            t.string('name',{nullable: true }),
              t.string('email'),
              t.string('password')
          }
        })
      },
  (...)</pre>
i have error:
image

How can i achieve such schema ?

tgriesser:
This is a big new feature for Nexus. It allows defining arbitrary new members to the definition blocks in a type-safe way.

```ts
t.relayConnection("members", {
  type: User,
  args: {
    limit: intArg(),
    offset: intArg()
  },
  edges(root, args, ctx, info) {
    return ctx.user.getMemberEdges(root.id, args)
  },
  pageInfo(root, args, ctx, info) {
    return ctx.user.getMemberEdgesPageInfo(root.id, args)
  },
});
```

See the example implementation of [relay connection](https://github.com/prisma/nexus/blob/dynamic-field/src/extensions/relayConnection.ts) and a more generic "[collection](https://github.com/prisma/nexus/blob/dynamic-field/src/extensions/collection.ts)" - both still may be refined a bit.

As a part of this, Nexus will now walk the leaf nodes when constructing the schema, meaning that not all types will need to be imported into the root - making schema construction significantly simpler and will automatically provide better interop for graphql-js (fixes #136, #70, #88).

Other changes include input types being able to be provided directly on the arg rather than requiring them to be wrapped in `arg()`, and InputObjectTypes having an `inputType.asArg(cfg)` method as a shorthand for `arg({type: inputType, ...cfg})`. Fixes #137

Fixes #22


tgriesser:
Just published this under `0.12.0-beta.1` if folks want to try it out

stalniy:
I’m a bit worried about performance because I plan to rebuild scheme in runtime. 

I did a quick measurement of a big amount types and nexus builder was like 4 times slower (current version, not beta) than regular `GraphQL` scheme builder. 

Now I see that you walk recursively over NamedTypes (regular graphql abstractions). `GraphQlSchema` does the same thing, so the question: 

Don’t you repeat the job which eventually will be done one more time by `GraphQLScheme`?  And if so, whether there is a chance not to walk at least over regular graphql types?

tgriesser:
> I’m a bit worried about performance because I plan to rebuild scheme in runtime.

Great point - I do want to continue evaluating the value of abstractions vs overhead. This is also just a rough first pass, I'm sure there's things that can be simplified.

> I did a quick measurement of a big amount types and nexus builder was like 4 times slower (current version, not beta) than regular GraphQL scheme builder.

Interesting, what were the numbers you were seeing? # of types and measurement numbers. What is the use case where you're rebuilding the schema in runtime?

> Don’t you repeat the job which eventually will be done one more time by GraphQLScheme?
> And if so, whether there is a chance not to walk at least over regular graphql types?

Yes, there is a bit of duplication here. In order for Nexus types to work in GraphQL Schema, all types first need to be made available as GraphQL types. Due to Nexus allowing string references to object types, all types need to be available by the time fields are generated, which means we need to walk the types a first time to ensure they're all available, and then a second to build them into GraphQL types. 

Thinking about it more though, perhaps this is an edge case and we can just say that mixing the string reference and type gathering from fields is not supported.

Anyway, I appreciate the concerns and am going to look more into how to best reconcile performance vs abstractions.

stalniy:
@tgriesser it's quite stupid performance measurement script :) I published it [here](https://codesandbox.io/s/nexusvsgraphqlperf-qwu4p)

I test 4 cases:
1. 1st time generate 10000 types in nexus
2. 2nd time generate 10000 types in nexus
3. generate 6000 types using `graphql` (query types are generated as plain objects and eventually added to schema as single type)
4. merge everything into single schema using `mergeSchemas` (from `graphql-tools`)

The results:

```
nexus schema 0-2000: 419.945ms
nexus schema 2001-4000: 259.854ms
native schema: 33.046ms
merge schemas 3 schemas: 961.564ms
```

So, the difference between `graphql` native types and nexus is 12 times. 500 ms may be a big bottleneck for Node.js app, that's why I'm concerned about performance

tgriesser:
Thanks for the numbers.

True, there is some overhead, as is the price paid with any abstraction. But this cost is typically only incurred on application startup which presumably in most applications is before an application begins serving any traffic and performance matters.

You mentioned wanting to rebuild the schema in runtime, what is your use case for that?

stalniy:
I want to allow users to define their own types. Based on that types I plan to rebuild graphql scheme. 

If there is no way to optimize abstraction then it would be good to be able to have an API which return graphql types from nexus types. So, I can cache them in memory and do “smart” scheme updates. 

So, instead of `makeScheme` I want to call something like `resolveTypes` what returns me `{ query, mutation, subscription, types }` and later I can create scheme using `new GraphQLScheme()`

stalniy:
Another point:
I’d like nexus to delegate as much as possible validation logic to `graphql` package in order to speed up scheme building process.

tgriesser:
> If there is no way to optimize abstraction then it would be good to have an API which just converts nexus types to graphql types. So, I can cache them in memory and do “smart” scheme updates.

> So, instead of makeScheme I want to call something like resolveTypes what returns me { query, mutation, subscription, types } and later I can create scheme using new GraphQLScheme()

> Another point: I’d like nexus to delegate as much as possible validation logic to graphql package in order to speed up scheme building process.

It sounds like for these specific goals/requirements, you'd be better off building your own abstraction on top of the graphql-js primitives. Or you can mix/match some of the internals as needed... PR's are also welcome if you find some good opportunities to optimize.

blocka:
Is there a reason that unions are not included here? https://github.com/graphql-nexus/schema/blob/a6c29bae643471b200a2e1e74d61afd99f54c466/src/builder.ts#L694

PR #144, opened 2019-05-20
Add scalar examplex (Upload, DateTime) for new users

homoky:
When I started with Nexus I had to find how to declare Upload and DateTime. It should be defined in the docs. Feel free to update it as you want.


tgriesser:
Sounds good, thanks!

PR #154, opened 2019-06-09
Fix mistakes in "Example Schema" of getting started

peterp:
Added forgotten `enumType` import and used `args` over deprecated `fieldArg`


tgriesser:
Thanks!

PR #174, opened 2019-07-04
Fix inline functions for custom scalars

estrada9166:
You can check the issue [here](https://nexus.js.org/playground).

```js
t.date('updatedAt', () => new Date());
``` 
is not working as expected, as it is right now it only works with:

```js
t.field('updatedAt', {
  type: 'Date',
  resolve: () => new Date()
});
```

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -1170,14 +1170,15 @@ export class SchemaBuilder {
   ) {
     // @ts-ignore
     block[methodName] = (fieldName: string, opts: any) => {
-      // @ts-ignore
-      const fieldConfig = {
+      let fieldConfig = {
         type: typeName,
-        ...opts
       }
 
       if (typeof opts === 'function') {
+        // @ts-ignore
         fieldConfig.resolve = opts
+      } else {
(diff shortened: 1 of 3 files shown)


tgriesser:
Ah great find - illustrates the danger of `// @ts-ignore` 😄

I don't believe there's any proper tests for the new dynamic field stuff yet, but would you be able to see if you could add one that covers this?

Also I assume the changes to `en.json` were autogenerated?

estrada9166:
I created one test, I hope it's the expected one. Also, I think I might need some help to fix the failing types

> Also I assume the changes to en.json were autogenerated?

Yes! it was autogenerated when I started the website.

tgriesser:
> I created one test, I hope it's the expected one. 

Yep, looks great!

> Also, I think I might need some help to fix the failing types.

Just need to add a `// @ts-ignore` above the line:

```ts
t.date('testDate', () => now)
```

in the example and that should do it - because the types for the schema aren't auto-inferred it doesn't know what the schema extensions are. Not sure there's a good way around this without having separate TypeScript projects for each test case.

PR #178, opened 2019-07-08
Add dynamicObjectProperty, fix #161

### LINKED ISSUE #161: Feature request: Add `dynamicObjectProperty`
## Description

Currently, it is only possible to have dynamic output **method**s as shown here:

https://github.com/prisma/nexus/blob/b1089a5665e5e3eaefdf1ffb4424119e1bb66828/src/builder.ts#L1143-L1150

For nexus-prisma, we need a way to have dynamic object properties (or some better naming), to allow the following pattern:

```ts
objectType({
  name: 'SomeType',
  definition(t) {
    t.model.something()
  }
})
```

👆On the example above, `model` is just a simple object property instead of an actual method.

I think it can be fairly easily implemented using a getter:

```ts
Object.defineProperty(block, methodName, {
  get() {
    return val.value.factory({
      typeDef: block,
      builder: this,
      typeName: block.typeName
    });
  },
  enumerable: true
});
```

tgriesser:
Fixes #161

@schickling I believe this is the last piece you need in order to get nexus-prisma back on the main library, correct?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ website/build
 examples/*/dist
 website/static/playground-dist
 yarn-error.log
+coverage/*
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -94,6 +94,7 @@ import {
   isNexusDynamicInputMethod,
   isNexusDynamicOutputMethod,
   isNexusArgDef,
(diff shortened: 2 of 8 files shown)


schickling:
This looks great @tgriesser! :shipit:

PR #196, opened 2019-08-12
Fix typo in README.md

hkjorgensen:
just a small typo


jasonkuhrt:
Hey @hkjorgensen, what about `permissions` seems to be a better term to you than `authorization`? Auth[z/n] aka. authorization & authentication is the generally the entrypoint terminology in the auth space.

hkjorgensen:
@jasonkuhrt that's a good suggestion, but I just took the headline from actual repository that it links to - feel free to change :-)

jasonkuhrt:
@hkjorgensen Ah! Thanks, it would be great if you could update the other doc actually 😂

I'll close this side as its the more accurate title I think.

PR #199, opened 2019-08-18
Fix #189, reference dist/index.d.ts in types

### LINKED ISSUE #189: Typescript import suggests `nexus/dist`
Hey folks!

Whenever I try to "auto import" in VS Code, it suggests imports from `nexus/dist` instead of `nexus`... Any idea why that is? If there's any more info I can provide to help debug this let me know.

Thanks!

tgriesser:
- References dist/index.d.ts in types, (fix #189)
- Adds [declarationMap](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#new---declarationmap) to reference the original source rather than the types file when jumping to type definition


.vscode/settings.json
```diff
@@ -5,8 +5,7 @@
     "**/.hg": true,
     "**/CVS": true,
     "**/.DS_Store": true,
-    "**/node_modules": true,
-    "**/dist/**": true
```
jasonkuhrt: This just hides it in the vscode file tree right?

tgriesser: Yep.

tgriesser:
> On the one hand it seems this is only meaningful so long as we publish nexus with the source code, e.g.:

Yes, I assumed that we would continue doing that. I personally enjoy when libraries link to the actual source code so I can easily drop a debugger in / see what's going on if I want to, though I could see the argument for only needing to care about the external API as well. Interested to hear others' thoughts on this.

jasonkuhrt:
> only meaningful so long as we publish nexus with the source code

Oh, but we do publish `src` so there is that benefit (whatever it is) too:

```bash
# some project
$ ls node_modules/nexus | grep src
src
```

jasonkuhrt:
> I personally enjoy when libraries link to the actual source code so I can easily drop a debugger in / see what's going on if I want to

Agree!

PR #205, opened 2019-08-23
fix: dynamic output property

jasonkuhrt:
This change addresses the last thing remaining in the way before `nexus-prisma` can get back on the nexus mainline.

https://github.com/prisma/nexus-prisma/issues/304

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -126,7 +126,10 @@ import {
   NexusExtendInputTypeConfig,
 } from "./definitions/extendInputType";
 import { DynamicInputMethodDef, DynamicOutputMethodDef } from "./dynamicMethod";
-import { DynamicOutputPropertyDef } from "./dynamicProperty";
+import {
+  DynamicOutputPropertyDef,
+  DynamicInputPropertyDef,
+} from "./dynamicProperty";
 
 export type Maybe<T> = T | null;
 
@@ -266,6 +269,11 @@ export type DynamicOutputProperties = Record<
   DynamicOutputPropertyDef<string>
 >;
 
+export type DynamicInputProperties = Record<
+  string,
+  DynamicInputPropertyDef<string>
+>;
+
 /**
  * Builds all of the types, properly accounts for any using "mix".
  * Since the enum types are resolved synchronously, these need to guard for
@@ -324,6 +332,11 @@ export class SchemaBuilder {
    * Add dynamic output properties
    */
   protected dynamicOutputProperties: DynamicOutputProperties = {};
+
+  /**
+   * Add dynamic input properties
+   */
+  protected dynamicInputProperties: DynamicInputProperties = {};
 
   /**
    * All types that need to be traversed for children types
@@ -1143,18 +1156,27 @@ export class SchemaBuilder {
     return obj;
   }
 
+  // MARK
   addDynamicInputFields(block: InputDefinitionBlock<any>, isList: boolean) {
     eachObj(this.dynamicInputFields, (val, methodName) => {
       if (typeof val === "string") {
         return this.addDynamicScalar(methodName, val, block);
       }
 
-      // At this point `methodName` identifier name can be a misnormer. If the
-      // given factory returns a function then methodName is accurate, but not
-      // if the factory returns something else. For example it may return an
-      // object of methods (aka. methods under a namespace/field/property)
-      // in which case a better identifier might be `namespace`, `propName`, ...
-      Object.defineProperty(block, methodName, {
+      // @ts-ignore
+      block[methodName] = (...args: any[]) => {
+        const config = isList ? [args[0], { list: isList, ...args[1] }] : args;
+        return val.value.factory({
+          args: config,
+          typeDef: block,
+          builder: this,
+          typeName: block.typeName,
+        });
+      };
+    });
+
+    eachObj(this.dynamicInputProperties, (val, propertyName) => {
+      Object.defineProperty(block, propertyName, {
         get() {
           return val.value.factory({
             args: [],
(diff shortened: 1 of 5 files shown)


tgriesser:
This is close, but not quite the change we want - we want the ability to define both dynamic input methods & properties. I know what the issue is here, ran into it when adding tests for #200 - will open a little later this morning.

jasonkuhrt:
> we want the ability to define both dynamic input methods & properties

Sounds good, but isn't that possible now, just depending on what `factory` returns? E.g.:

```ts
const block: any = {}

Object.defineProperty(block, "foo", {
  get: () => {
    return (): string => {
      return "bar"
    }
  },
})

Object.defineProperty(block, "nested", {
  get: () => {
    return {
      foo: (): string => {
        return "bar"
      },
    }
  },
})

console.log(block.foo())
console.log(block.nested.foo())
```

?

src/builder.ts
```diff
@@ -1149,15 +1149,17 @@ export class SchemaBuilder {
...
-      block[methodName] = (...args: any[]) => {
-        const config = isList ? [args[0], { list: isList, ...args[1] }] : args;
-        return val.value.factory({
-          args: config,
-          typeDef: block,
-          builder: this,
-          typeName: block.typeName,
-        });
-      };
+      Object.defineProperty(block, methodName, {
```
jasonkuhrt: `methodName` is a bit of a misnomer since it can be a field, just depends on what `factor` is returning. Maybe call this `methodOrFieldName` for example.

jasonkuhrt: Addressed as comment

src/builder.ts
```diff
@@ -1148,16 +1148,23 @@ export class SchemaBuilder {
       if (typeof val === "string") {
         return this.addDynamicScalar(methodName, val, block);
       }
-      // @ts-ignore
-      block[methodName] = (...args: any[]) => {
-        const config = isList ? [args[0], { list: isList, ...args[1] }] : args;
```
jasonkuhrt: I need to review this existing `args` logic more and see what's being lost in this PR currently. Pretty sure something.

tgriesser:
This PR originally added `dynamicObjectProperty` - https://github.com/prisma/nexus/pull/178 the intention was that we should be explicit as to whether it's adding a property or a method rather than overloading it.

If we want to keep them as distinct concepts, I believe what's left is: 
1. adding the equivalent for "dynamic properties" on input types
2. fixing quite a few places the symbols I added were incorrect since there weren't any tests and they needed to be cleaned up: 

https://github.com/prisma/nexus/pull/200/files#diff-84231b52da5756993249faf30a25c7c3R34
https://github.com/prisma/nexus/pull/200/files#diff-629f635a7466a97bd69d407fe79be89aR150
https://github.com/prisma/nexus/pull/200/files#diff-cf514b96691f8608ebc06c9e05285e58R66
https://github.com/prisma/nexus/pull/200/files#diff-b2957fb214e0db032454663585fcdd3bR32
https://github.com/prisma/nexus/pull/200/files#diff-629f635a7466a97bd69d407fe79be89aR166

I had also written some tests for the "dynamic" method/property pieces on that plugin branch that might be useful to port over:

https://github.com/prisma/nexus/pull/200/files#diff-ce313718055422103750b425fdbcb1c2R1

Let me know what you think of this approach / if this clears up what I meant by ^.

jasonkuhrt:
@tgriesser thanks for the info!

> the intention was that we should be explicit as to whether it's adding a property or a method rather than overloading it.

Sounds good, let's stick to that.

> adding the equivalent for "dynamic properties" on input types

I noticed:

```ts
import { DynamicInputMethodDef, DynamicOutputMethodDef } from "./dynamicMethod";
import { DynamicOutputPropertyDef } from "./dynamicProperty";
```

IIUC, one this PR is done, we'll have:

```ts
import { DynamicInputMethodDef, DynamicOutputMethodDef } from "./dynamicMethod";
import { DynamicInputPropertyDef, DynamicOutputPropertyDef } from "./dynamicProperty";
```

> I had also written some tests for the "dynamic" method/property pieces on that plugin branch that might be useful to port over:

Awesome, will look into doing this 👍

> Let me know what you think of this approach / if this clears up what I meant by ^.

Yes helps a lot! I'm still vague on the different between input and output types and more so on `the symbols I added` (IIUC the `withNexusSymbol(...)` stuff). But I'll continue to internalize this stuff and see if I can get all the way or at least to a meaningful iteration that you can provide further feedback upon 🙏 :)

jasonkuhrt:
Hey @tgriesser I started work on adding `dynamicInputProperty` but soon realized that this is, at least presently, not the minimal step to take. Doing so would be adding a feature. And nexus-prisma does not actually use `dynamicInputProperty`. It is actually interested in `dynamicOutputProperty`. It _currently_ (via the nexus fork `@prisma/nexus`) uses `dynamicOutputMethod`. So two things need to happen that I can see:

1. Update `nexus-prisma` to use `dynamicOutputProperty`
2. Fix `dynamicOutputProperty` in nexus :) 

This PR is now focused on #2 above.

From my quick local tests I was able to get tests ~passing on nexus-prisma with the changes herein. Unfortunately linking nexus into nexus-prisma runs into trouble regarding different versions of `graphql` from different "realms":

```
    Cannot use GraphQLDirective "@include" from another module or realm.

    Ensure that there is only one instance of "graphql" in the node_modules
    directory. If different versions of "graphql" are the dependencies of other
    relied on modules, use "resolutions" to ensure only one version is installed.

    https://yarnpkg.com/en/docs/selective-version-resolutions

    Duplicate "graphql" modules cannot be used at the same time since different
    versions may have different capabilities and behavior. The data from one
    version used in the function from another could produce confusing and
    spurious results.
```

Any idea how to get around that? Anyways one use-case for ci-cd lib publishing on every pr commit :)

jasonkuhrt:
I tried pulling in some of the tests from #200 but they were non-trivial as they depended upon numerous other additions in that PR not present here. Therefore stopping trying to port any over here now.

On account of this PR being now only additive and rather simple and clear, I am going to move ahead with it.

PR #217, opened 2019-09-09
Correct error message about absolute paths

mssngr:
The absolute path error was driving me nuts, as I thought it was telling me my string was not a string. I've updated the error message to correctly communicate the error.


tgriesser:
Ahhh definitely an oversight, thanks @mssngr!

PR #227, opened 2019-09-23
Remove file:. from devDependencies

tgriesser:
This was really slowing down yarn install locally, because it seemed to be linking everything in `node_modules` recursively or something. This taps into the `formatTypegen` method to accomplish the same end-goal, testing the local `nexus` in the integration tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "dev": "yarn link-examples && tsc -w",
     "test": "jest",
-    "test:ci": "yarn add file:. && jest -i --coverage",
+    "test:ci": "jest -i --coverage",
     "build": "tsc",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -50,7 +50,6 @@
     "graphql-iso-date": "^3.6.1",
     "jest-watch-typeahead": "^0.3.1",
     "lint-staged": "^7.3.0",
-    "nexus": "file:.",
```
jasonkuhrt: What was this even achieving, I forget now.

tgriesser: Making the import of `import { core } from "nexus"` work here:

https://github.com/prisma-labs/nexus/pull/227/files#diff-2d9f20d621ccd8f7a36d3d1bb829f8e8

PR #230, opened 2019-09-26
feat: by default typegen as an @types package

### LINKED ISSUE #209: Make schema output with a good defaults
It would be nice to provide good defaults for `outputs` in `makeSchema`. With this `makeSchema` would become a one-liner. Nice default would be one provided in `README`.

Before:
```js
const schema = makeSchema({
  types: [Query],
  outputs: {
    schema: __dirname + "/generated/schema.graphql",
    typegen: __dirname + "/generated/typings.ts",
  },
});
```
After:
```js
const schema = makeSchema({
  types: [Query],
});
```

jasonkuhrt:
* Centralizes and formalizes transformation from outward facing options to inward facing config
* Increases expressivity of outputs config
* adds defaults for outputs

closes #209


tests/__snapshots__/typegen.spec.ts.snap
```diff
@@ -148,7 +148,7 @@ exports[`typegen should print the full output 1`] = `
  * Do not make changes to this file directly
  */
 
-import * as t from \\"./_helpers\\"
+import * as t from \\"./tests/_helpers\\"
```
jasonkuhrt: I don't understand yet why this changed.

jasonkuhrt: Tests against examples are also failed because of this. The SW API.

jasonkuhrt:
Relevant to this issue is https://github.com/prisma-labs/nexus-prisma/issues/430 where the typegen-in-node-modules strategy (nexus-prisma) is not working out in some scenario. I'm hoping to debug more with @iherger because we this direction is great DX hence default so long as we work out the kinks.

src/builder.ts
```diff
@@ -170,24 +172,41 @@ export const UNKNOWN_TYPE_SCALAR = decorateType(
...
-         * Absolute path where the GraphQL IDL file should be written
+         * TypeScript declaration file generation settings. This file
+         * contains types reflected off your source code. It is how
+         * Nexus imbues dynamic code with static guarnatees.
+         *
+         * Defaults to being enabled when `process.env.NODE_ENV !== "production"`.
+         * Set to true to enable and emit into default path (see below).
+         * Set to false to disable. Set to a string to specify absolute path.
+         *
+         * The default path is node_modules/@types/__nexus-typegen__core/index.d.ts.
```
tgriesser: I recently did a deep-dive in the rules of type resolution but can't remember now, whether TS picks up `.ts` as well as `.d.ts` files by default in `@types`. 

I tend to use `.ts` files for the nexus gen in my own projects, because I often have `"skipLibCheck": true,` set in `tsconfig` (to suppress extraneous errors in 3rd party dependencies). When this is set, any type errors in all `.d.ts` files are silently ignored, which can be a gotcha.

src/builder.ts
```diff
@@ -354,16 +450,16 @@ export class SchemaBuilder {
    */
   protected finalized: boolean = false;
 
-  constructor(protected config: BuilderConfig) {
+  constructor(protected options: BuilderOptions) {
```
tgriesser: Minor, but can we stay consistent w/ `Options` vs `Config`? I think everywhere should be `Config`, and if we need to change it, we could just make the config more explicit, e.g. `BuilderInternalConfig` - unless there's a reason to do some of each?

src/typegenMetadata.ts
```diff
@@ -21,40 +20,28 @@ import {
...
-      }
-      if (this.typegenFile) {
-        const value = await this.generateTypesFile(
-          sortedSchema,
-          schema.extensions.nexus
-        );
-        await this.writeFile("types", value, this.typegenFile);
-      }
+    if (this.config.outputs.schema) {
+      await this.writeFile(
```
tgriesser: I think @Weakky mentioned wanting to change from `writeFile` to `unlink / writeFile`, apparently vscode picks up the changes better this way if the file is replaced rather than modified? Not sure if we should do that as part of this or in a separate PR.

jasonkuhrt: @tgriesser sounds great, but yeah separate PR 👍

src/builder.ts
```diff
@@ -245,15 +273,83 @@ export interface TypegenInfo {
...
+export interface BuilderConfig extends BuilderOptions {
+  outputs: {
+    schema: false | string;
+    typegen: false | string;
+  };
+}
+
+function resolveBuilderOptions(config: BuilderOptions): BuilderConfig {
+  if (
+    typeof config.outputs === "object" &&
```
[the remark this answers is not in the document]
jasonkuhrt: Yeah thought about it, agree. Thanks for poking. We just have to account for the fact that `config.outputs === false` is not the same semantic here as `config.outputs === undefined`.

src/builder.ts
jasonkuhrt: > When this is set, any type errors in all .d.ts files are silently ignored, which can be a gotcha.

I just tried `skipLibCheck` on the nexus-prisma example app (which generates into `@types`) and got this result: 

image

Did I run my test incorrectly? Or did I misunderstand your point?

src/builder.ts
jasonkuhrt: So the underlying idea is that at the top layer is some human-centric settings that permit nullables, shorthands, etc. Once these enter through the library's entrypoint(s) though, these settings get "resolved", short hands processed, defaults supplied, etc.

To make this idea more readable in the codebase, I thought of using different terms. `options` being the higher layer, and `config` being the lower, resolved, layer.

How do you feel about that?

Another point might be that the idea was rolled out semi-spuriously, rather than uniformly across the codebase.

src/builder.ts
```diff
@@ -245,15 +273,84 @@ export interface TypegenInfo {
...
+    schema: false | string;
+    typegen: false | string;
+  };
+}
+
+function resolveBuilderOptions(config: BuilderOptions): BuilderConfig {
+  if (
+    config.outputs !== undefined &&
+    typeof config.outputs !== "boolean" &&
+    typeof config.outputs.schema === "string"
```
jasonkuhrt: @tgriesser My revised version is still not safe in JS. This line could still be reached by `null`, in JS. I'm hesitant to starting coding type checks optimized for JS though. If they want great feedback on type errors, don't use JS... For us, I doubt we're  going to have consistent and uniform tight type checks everywhere for JS, so why bother with sprinkles?

tgriesser: Yeah, actually `null` was the only thing I was suggesting we guard against, from the original code, it would have been:

```ts
if (
  config.outputs &&
  typeof config.outputs === "object" &&
  typeof config.outputs.schema === "string"
)
```

> For us, I doubt we're going to have consistent airtight type checks everywhere for JS, so why bother with sprinkles of it?

I agree with this, I'm fine with leaving it out - the `null` / `typeof` is one I still generally try to guard against just because it can be a confusing runtime error, `cannot read property schema of null`.

src/builder.ts
tgriesser: Point was if there was a broken import or custom type or something in the `.d.ts`, I believe it would just turn the error into an `any` rather than actually reporting the error.

jasonkuhrt: Ah gotchya. And yeah, catching up to you here, confirmed.

If I attempt to use `index.ts` instead of `index.d.ts` in `@types` packages I get a compile error `error TS2688: Cannot find type definition file for '__nexus-typegen__nexus-core'.`.

I will do a bit of research to see if there is any solution that would permits the benefits of using `@types` without succumbing to the effects of `skipLibCheck`.

jasonkuhrt: So did not find a solution; thoughts:

- A baseline point, obviously it is not ideal to use `skipLibCheck`. Two use-cases originally IIUC were 1) work around incompatible dep/transient dep typing integration issues; 2) compile perf. With modern TS incremental build stuff, I would hope perf use-case is gone/going.
- IIUC/Question: it should be impossible for user code to introduce bug in typegen
- When user _needs_ `skipLibCheck`, they still benefit from typegen in `@types`, so long as the typegen is bug free
- Considering the above points in sum, IIUC, using `@types` as default is problematic only when the stars align thusly: 1) nexus lib has a typegen bug 2) user hits the bug path 3) user _needs_ `skipLibCheck`.
- Escape hatch solution: user manually configures typegen paths

I think such a tradeoff would be reasonable.

jasonkuhrt: I found a discussion for a granular version of `skipLibCheck` here https://github.com/microsoft/TypeScript/issues/30511. If TS had this feature, it would further reduce pressure against nexus type in `@types` being a good default, because some amount of current `skipLibCheck` cases would be able to adjust to the granular form.

I will add our use-case to the discussion over there.

Weakky: > Point was if there was a broken import or custom type or something in the .d.ts, I believe it would just turn the error into an any rather than actually reporting the error.

Would it still be the case with `noImplicitAny: true` ?

I generally agree with your point @jasonkuhrt. The whole point of such defaults is to reduce friction for newcomers. I'm still a bit concerned by the issue cited above though. Broken import can happen easily when messing with `typegenAutoConfig`, and inputting some wrong file paths

jasonkuhrt: > noImplicitAny: true

Yep still the same.

> Broken import can happen easily when messing with typegenAutoConfig (to enhance the backing types), and inputting some wrong file paths

Ah, thanks for clarifying that. Agree that increases the downside potential.

Still feel it is the right tradeoff.

We probably want good user docs for this part of the system. User awareness can do its part to help narrow the unhappy path.

src/builder.ts
jasonkuhrt: I will take a different tact to reduce the diff here, that being, `config / Config` and `internalConfig / InternalConfig`.

jasonkuhrt: Done, now `options` does not show up anywhere, just `config` and variants of that.

jasonkuhrt:
I've integrated the general feedback. The only thing left is to align on currently is the above unresolved thread. I don't see an alternative solution to consider, and I think the changes are overall sound.

src/builder.ts
tgriesser: Yeah I think what you have as the default is reasonable. Let's do it.

PR #234, opened 2019-09-30
refactor: thread schema config through build types

jasonkuhrt:
Simplify data flow by threading schema config further.

This is technically a breaking change since `buildTypes` was part of the exported API. However it seems to be an internal that shouldn't really be considered part of the public semver contract?


jasonkuhrt:
Might revisit this later but not worth it right now.

PR #235, opened 2019-09-30
chore: dev script for lib work

jasonkuhrt:


package.json
```diff
@@ -7,10 +7,11 @@
   "description": "Scalable, strongly typed GraphQL schema development",
   "homepage": "https://nexus.js.org",
   "scripts": {
-    "dev": "yarn link-examples && tsc -w",
+    "website:dev": "yarn link-examples && tsc -w",
```
Weakky: While `gulp link-examples` is run from the `website` directory, `link-examples` link all the nexus examples to the local nexus, so `website:dev` doesn't seem to be the right name. Still think it's valuable to have a command that doesn't run all the linking before linking tsc

Weakky: Maybe we can do something similar to `lerna` ? Have `website:dev` be named `bootstrap` ?

jasonkuhrt: Not sure I follow. I simply want to make `dev` for library dev rather than website dev. This comes in hand when, for example, using yarn link workflows across a few projects locally.

tgriesser: Yeah I think `dev:examples` may be a better name for this.

PR #236, opened 2019-09-30
feat(plugins): onInstall hook

jasonkuhrt:
#### TODO
- [x] tests
- [x] runtime validation on 1) the handler type 2) the handler return value.


src/builder.ts
```diff
@@ -1330,6 +1336,9 @@ export function buildTypes<
   schemaBuilder?: SchemaBuilder
 ): BuildTypes<TypeMapDefs> {
   const builder = schemaBuilder || new SchemaBuilder(config);
+  (config.plugins || [])
+    .filter((p) => p.onBuild)
+    .forEach((p) => () => p.onBuild!(types, builder));
```
tgriesser: i’m not sure I see the value of being able to access the `types` at that stage, as it’s an incomplete representation of the total types, as opposed to adding some public helper methods like `.hasType(typeName)` to `builder`, and then calling `onBeforeBuild` in this line:

https://github.com/prisma-labs/nexus/blob/f37f20b20631931382ed12d61c5173a2e24e387d/src/builder.ts#L491-L495

Directly after `walkTypes`.

src/builder.ts
```diff
@@ -170,7 +170,13 @@ export const UNKNOWN_TYPE_SCALAR = decorateType(
   }
 );
 
+export interface Plugin {
+  // TODO TS 3.6 Generator type
+  onBuild?: (types: any, builder: SchemaBuilder) => Generator; //Generator<void, void, void>;
```
tgriesser: Did we need a generator for some reason?

jasonkuhrt: Nope could not. If we discover need for `onAfterBuild` we can revisit in future I guess.

src/builder.ts
```diff
@@ -366,6 +381,15 @@ export class SchemaBuilder {
...
   }
 
+  // /**
+  //  * Find out if a type has been defined or not, and if it has been, return it.
+  //  * This is useful, for example, for plugins that want to somehow adapt to the
+  //  * user's type defs.
+  //  */
+  // getNamedType(name: string): null | GraphQLNamedType {
+  //   return this.finalTypeMap[name] || null;
+  // }
```
jasonkuhrt: Remove before PR merge. This was me working through approach where plugins use builder.

src/builder.ts
```diff
@@ -275,6 +280,16 @@ export type DynamicOutputProperties = Record<
   DynamicOutputPropertyDef<string>
 >;
 
+export type TypeDef = AllNexusNamedTypeDefs | GraphQLNamedType;
+
+export type BlockDef =
```
jasonkuhrt: Seems correct to me. Nexus works with a superset concept of blocks of which type definitions are a subset.

src/builder.ts
```diff
@@ -1330,33 +1345,63 @@ export function buildTypes<
...
+    .filter((p) => p.onBeforeBuild)
+    .reduce((blocksAcc, p) => p.onBeforeBuild!(blocksAcc), blocks);
+  pluggedBlocks.forEach(builder.addType);
   return builder.getFinalTypeMap();
 }
 
-function addTypes(builder: SchemaBuilder, types: any) {
-  if (!types) {
-    return;
+function unwrapBlocks(blocks: any, builder: SchemaBuilder): BlockDef[] {
```
jasonkuhrt: Separating unwrap from builder so that plugins have chance to cut in and do work without having to unwrap themselves.

src/builder.ts
```diff
@@ -1330,33 +1345,63 @@ export function buildTypes<
...
+
+function unwrapBlocksDo(
+  acc: BlockDef[],
+  givenBlockDefs: object,
+  builder: SchemaBuilder
+): BlockDef[] {
+  // NOTE We mutate `acc` over purity for performance.
+
+  if (isNexusWrappedType(givenBlockDefs)) {
+    acc.push(givenBlockDefs.fn(builder));
```
jasonkuhrt: Not sure about this part.

src/builder.ts
tgriesser: I think you might be missing what I meant in the other comment above. The version you pushed still doesn't do the same thing as `builder.walkTypes()`, so it would not be useful to call `onBeforeBuild` at this stage because it can't know what all of the types until they are fully unwrapped.

I appreciate the desire to make things more pure from a data perspective, but ultimately I think it will be simpler due to the self-referential property of the types to keep all of this internal in `builder`, as long as we assume that the builder is effectively a protocol for adding the types and calling `getFinalTypeMap` reifies the provided state into a shape that can be provided to the schema.

It might be better to expose what we want to be available from the builder via an explicit interface, rather than the `builder` itsself, sort of what the definition blocks do in the `walkInputType`, `walkOutputType`:

```ts
onBeforeBuild({
  hasType(typeName: string): boolean
  addType(blockDef: BlockDef): void
})
```

jasonkuhrt: Yeah these changes came while your first round of feedback dropped and I just completed the iteration anyways. Point received!

src/plugins.ts
```diff
@@ -0,0 +1,16 @@
...
+
+type Hooks = {
+  onInstall: OnInstallHook;
+};
+
+type OnInstallHook = (
+  onInstallHandler: (builder: BuilderFacade) => { types: any[] }
+) => void;
+
+export { OnInstallHook, Plugin, Hooks };
```
jasonkuhrt: How do you feel about batch export at end of file? I like the symmetry with batch import at beginning of file. Also less clutter I think. Colocating the answer to "is this exported" is lost though. Arguably IDE outline view etc. can support but not enough I guess.

src/plugins.ts
```diff
@@ -0,0 +1,16 @@
...
+};
+
+type Plugin = (hooks: Hooks) => void;
+
+type Hooks = {
+  onInstall: OnInstallHook;
+};
+
+type OnInstallHook = (
+  onInstallHandler: (builder: BuilderFacade) => { types: any[] }
```
jasonkuhrt: I'm proposing here that the plugin hook handler uses `return` values to "write" back to nexus core, and keeps the passed builder facade argument for read-only operations. I think this lends to a data flow that is easier to understand. The plugin builds up types however it wants and _returns_ them to nexus core. Seems natural this way.

src/plugins.ts
```diff
@@ -0,0 +1,95 @@
...
+type BuilderFacade = {
+  hasType: (typeName: string) => boolean;
+};
+
+/**
+ * This is the Neuxs Plugin interface that allows users to extend Nexus at
+ * particular extension points. Plugins are just functions that receive hooks
+ * which can then be registered upon with callbacks.
+ */
+export type Plugin = (hooks: Hooks) => void;
```
jasonkuhrt: I am wavering between `hooks` and `lifecycle`.

src/plugins.ts
```diff
@@ -0,0 +1,95 @@
...
+
+type Hooks = {
```
jasonkuhrt: I am wavering between `hooks` and `lifecycle`.

jasonkuhrt:
I will write tests tomorrow morning.

src/plugins.ts
```diff
@@ -0,0 +1,95 @@
...
+      state.onInstallHandler = handler;
+    },
+  });
+
+  return {
+    triggerOnInstall: () => {
+      const { types } = state.onInstallHandler({
+        hasType: builder.hasType,
+      });
+      types.forEach(builder.addType);
```
jasonkuhrt: We should probably put some runtime checks here for better error handling. Especially helpful for JS.

src/plugins.ts
```diff
@@ -0,0 +1,95 @@
...
+  onInstall: OnInstallHook;
+};
+
+type OnInstallHook = (onInstallHandler: OnInstallHandler) => void;
+
+type OnInstallHandler = (
+  builder: BuilderFacade
+) => {
+  types: NexusAcceptedTypeDef[];
+};
```
jasonkuhrt: I intentionally made this 1) extensible and 2) strict.

By returning a record we allow for `onInstall` to do more things in the future. We can deprecate `types` key in favour of new ones that separate concepts, or whatever. Who knows.

Currently the only reason to use `onInstall` hook is to contribute type definitions. Rather than taking an optional approach and thus opening user up to error (e.g. accidental non-return) we just enforce it. We can easily change our minds later, relaxing the requirement, because turning this into a nullable would be backwards compatible.

src/plugins.ts
jasonkuhrt: We might want to use runtime DSLs that permit deriving (inferring) TS static types:

* https://github.com/gcanti/io-ts
* https://github.com/jquense/yup
* https://github.com/pelotom/runtypes

Normally this approach would be used for the IO boundary of an app. But we can kind of think of JS Nexus users as being in this camp too, where all data they are passing into Nexus API requires runtime validation.

I would prefer if we could codegen runtime checks from our TS types but don't know of a good tool for that. Definitely seems the runtime-to-inference is the dominant approach. Fair since that approach requires zero special tooling.

jasonkuhrt:
For interface consideration. Here is how nexus-prisma case is looking.

```ts
/**
 * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
 */
export const create = (options: Options): Plugin => {
  return hooks => {
    hooks.onInstall(nexusBuilder => {
      return {
        types: Builder.build({ ...options, nexusBuilder }),
      }
    })
  }
}
```

#### Alternative Styles

`s/hooks/lifecycle`:

```ts
/**
 * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
 */
export const create = (options: Options): Plugin => {
  return lifecycle => {
    lifecycle.onInstall(nexusBuilder => {
      return {
        types: Builder.build({ ...options, nexusBuilder }),
      }
    })
  }
}
```

Imperative style write-capable `nexusBuilder` facade:

```ts
/**
 * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
 */
export const create = (options: Options): Plugin => {
  return lifecycle => {
    lifecycle.onInstall(nexusBuilder => {
      Builder.build({ ...options, nexusBuilder })
    })
  }
}
```

Class Based:

```ts
/**
 * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
 */
export class NexusPrismaPlugin extends NexusPlugin {
  onInstall (nexusBuilder) {
    return {
        types: Builder.build({ ...options, nexusBuilder }),
      }
  }
}
```

Class Based + Imperative style write-capable `nexusBuilder` facade:

```ts
/**
 * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
 */
export class NexusPrismaPlugin extends NexusPlugin {
  onInstall (nexusBuilder) {
    Builder.build({ ...options, nexusBuilder })
  }
}
```

jasonkuhrt:
I think the POJO approach is better. I do not see the benefit of a closure that receives `lifecycle`. It just makes things a bit more complicated I feel.

jasonkuhrt:
@tgriesser happy with how this is looking, and IIRC it fully aligns with your original gist https://github.com/prisma-labs/nexus-prisma/pull/434/files#diff-f41e9d04a45c83f3b6f6e630f10117feR10-R18. So going to mark the above "alternatives" comment as resolved. Feel free to reopen the discussion though!

jasonkuhrt:
Milestone reached on [sister `nexus-prisma` pr](https://github.com/prisma-labs/nexus-prisma/pull/434), fully working atop this version of Nexus.

src/plugins.ts
tgriesser: A goal I had with nexus was to have `graphql-js` as the only dependency of the lib  if possible (other than `tslib`, for obvious reasons). This is also the point of `addTypes`, the predicate checks are meant to guard here. 

This is also why I recommended having a container class for `Plugin`, following the patterns elsewhere so we can easily add a predicate guard.

src/builder.ts
```diff
@@ -281,37 +293,39 @@ export type DynamicOutputProperties = Record<
...
+  protected allTypes: Index<NexusAcceptedTypeDef> = {};
   /**
    * Used to check for circular references.
    */
   protected buildingTypes = new Set();
   /**
    * The "final type" map contains all types as they are built.
    */
-  protected finalTypeMap: Record<string, GraphQLNamedType> = {};
+  protected finalTypeMap: Index<GraphQLNamedType> = {};
```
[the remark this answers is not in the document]
jasonkuhrt: Record generally means to me `struct` in Go or `object` in JS or `record` in Haskell. Data bag. `map` is overloaded, e.g. `map` function, concept of mappings, etc. I see `Index` being a derived 0(1) lookup cache. And I find that a distinct way to read this code. Another step to me would renaming `*Map` idents to `*Index`.

I don't `Record` is a poor name for lookups. I haven't seen such a pairing of terms outside the TS community. I don't personally find reading Record<string,` over and over the codebase idiomatic.

src/builder.ts
```diff
@@ -366,25 +380,21 @@ export class SchemaBuilder {
     return this.config;
   }
 
+  hasType = (typeName: string): boolean => {
+    return Boolean(this.allTypes[typeName]);
```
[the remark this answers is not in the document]
jasonkuhrt: Didn't form a strong opinion on dynamics here. But I agree now, hasType should be able ***GraphQL*** types, not other special nexus constructs.

Would `Boolean(this.finalTypeMap[typeName] || this.pendingTypeMap[typeName])` account for scalars?

src/plugins.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Create a Nexus Plugin. This function is just a convenience for not having to
+ * import the Plugin type. Feel free to do this instead:
+ *
+ *     import { Plugin } from "nexus"
+ *     export default { ... } as Plugin
+ *
+ */
+export const create = (plugin: Plugin): Plugin => plugin;
```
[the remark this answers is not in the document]
jasonkuhrt: > Could we rename this to plugin or nexusPlugin. I think we prefer declarative rather than imperative naming for our definitions...

Its so-named at the top left export for users. But Internally the idea was naming in accordance with the namespace of the module. Does that move you at all?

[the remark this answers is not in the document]
jasonkuhrt: > so we can detect cases where a plugin is perhaps imported in the types: [...] by mistake and not provided to plugins: [...].

IIUC This would be in an effort to support feedback for JS devs right?

While we could achieve similar with duck typing that would still violate your second point:

> just for clarity/consistency w/ existing patterns

👍

src/utils.ts
```diff
@@ -236,3 +236,17 @@ export function relativePathTo(
...
+ * indexed by some property, e.g. a list of users indexed by email.
+ */
+export type Index<T> = Record<string, T>;
+
+/**
+ * Returns its input. For example:
+ *
+ *     id(1) // -> 1
+ */
+export const id = <T extends unknown>(x: T): T => x;
```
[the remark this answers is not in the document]
jasonkuhrt: I had used it to export the `create` function `const export create = <Plugin>id` but at some point I was getting a TS error I didn't recognize (`unknown doesn't overlap ...`) so I just ditched it.

src/builder.ts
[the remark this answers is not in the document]
jasonkuhrt: Anyway its an orthogonal change to the PR, we can defer my points here, I'll revert these.

src/plugins.ts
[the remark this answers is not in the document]
jasonkuhrt: While `plugin` func is a fine middle ground, I did like having the plugins be as simple as a POJO that you could type with simply a structural type `Plugin`.

src/builder.ts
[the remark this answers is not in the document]
jasonkuhrt: I have made a change so that I think only GraphQL kind types will show up in the `allTypeDefs`.

src/plugins.ts
jasonkuhrt: With #242 in sight, I retract my proposition : )

jasonkuhrt:
Hey @tgriesser I think I've responded to all your feedback. There is at least one needing your follow up (the `allTypeDefs` data).

src/plugins.ts
jasonkuhrt: We agreed about this on slack chat.

PR #240, opened 2019-10-02
chore: simplify tsconfig

jasonkuhrt:
On my machine I get jest test matcher type errors. This PR fixes that. Its also just simpler and what I would expect in the first place.

Unless the original motivation for how it was remains, we should simplify.


jasonkuhrt:
Example: 

image

tgriesser:
This change was intentional, but I can figure out why it was breaking in your branch later.

PR #244, opened 2019-10-04
feat(config): env var for should generate artifacts

### LINKED ISSUE #243: SHOULD_GENERATE_ARTIFACTS environment variable
In discussing with @iherger about https://github.com/prisma-labs/nexus-prisma/issues/430 I realized that users need for a way to explicitly control `config.should_generate_artifacts` from the environment.

1. Typegen is derived data so not committed to the codebase.
2. In the `heroku` system, the build occurs as part of their multi-step deploy pipeline (it runs `npm run build`)
3. In the deploy pipeline `NODE_ENV` is set to `production` (or @iherger did you set that manually? Anyway doesn't really change the point)
4. The app fails to build because TS type checks fail because typegen is not present becuase `NODE_ENV` is `production`.

I propose that we build in support for allowing an environment variable to enable artifact generation. This way Heroku users could for example:

```json5
// package.json
"scripts": {
  "build": "NEXUS_SHOULD_GENERATE_ARTIFACTS=true ts-node src/schema.ts"
```

This would override the check against `NODE_ENV`. It would not override an explicit code setting of `config.should_generate_artifacts === false`.

jasonkuhrt:
Allow env var `NEXUS_SHOULD_GENERATE_ARTIFACTS` to configure `config.shouldGenerateArtifacts`.

This allows decoupling NODE_ENV from artifact gen config in a flexible way.

In particular this assists managing artifact gen in ci/cd workflows, sans source code changes.

closes #243


src/builder.ts
```diff
@@ -128,6 +128,7 @@ import {
 import { DynamicInputMethodDef, DynamicOutputMethodDef } from "./dynamicMethod";
 import { DynamicOutputPropertyDef } from "./dynamicProperty";
 import { decorateType } from "./definitions/decorateType";
+import { cloneDeep } from "lodash";
```
tgriesser: Let's not add lodash as a dependency 😄

jasonkuhrt: @tgriesser yes yes! :D I was just debugging something, on a wip that was doing deep clone stuff. All gone now though. 🙏

jasonkuhrt:
@tgriesser rfr!

tests/builder.spec.ts
```diff
@@ -4,83 +4,107 @@ import {
...
+      expect(internalConfig.outputs).toEqual(expected.outputs);
+    }
+  ); // prettier-ignore
+});
 
-  it.each(genOn)("with NODE_ENV=development %j", (given, expected) => {
-    process.env.NODE_ENV = "development";
-    expect(resolveBuilderConfig(given).outputs).toEqual(expected.outputs);
-  });
+describe("Environment variable influence over builder config shouldGenerateArtifacts", () => {
```
jasonkuhrt: Idea here is grouping tests pertaining to env var config.

jasonkuhrt:
Comments addressed. This PR looks tame, going to move forward with it.

PR #248, opened 2019-10-05
feat(typegen): nicer generated package name

jasonkuhrt:
This can impact the end-user when they, for example, are using the `types` TS compiler option. In that situation, expressing a whitelist of TypeScript type-packages, in a Nexus app, it would be important that the user also include typegen or else TSC would stop seeing it (forcing the user to provide custom typegen config). To make it easier for users to do that, let them refer to a pretty name, rather than something hard to predict, remember, and looks ugly.

#### TODO

- [x] update tests

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -303,7 +303,7 @@ export function resolveBuilderConfig(
   const defaultSchemaPath = path.join(process.cwd(), "schema.graphql");
   const defaultTypesPath = path.join(
     __dirname,
-    "../node_modules/@types/__nexus-typegen__core/index.d.ts"
+    "../node_modules/@types/nexus-typegen/index.d.ts"
   );
   const defaults = {
     outputs: {
--- a/tests/builder.spec.ts
+++ b/tests/builder.spec.ts
@@ -8,7 +8,7 @@ import { restoreEnvAfterEach } from "../src/core";
 
 const relativePath = (...paths: string[]): string => Path.join(__dirname, ...paths); // prettier-ignore
 const atTypesPath = relativePath("../node_modules/@types");
-const typegenDefault = Path.join(atTypesPath, "/__nexus-typegen__core/index.d.ts"); // prettier-ignore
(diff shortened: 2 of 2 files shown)


src/builder.ts
```diff
@@ -303,7 +303,7 @@ export function resolveBuilderConfig(
   const defaultSchemaPath = path.join(process.cwd(), "schema.graphql");
   const defaultTypesPath = path.join(
     __dirname,
-    "../node_modules/@types/__nexus-typegen__core/index.d.ts"
+    "../node_modules/@types/nexus-typegen/index.d.ts"
```
jasonkuhrt: 1. Is this name sufficiently risk-free of colliding with an actual future package called `nexus-typegen`?
2. In the future plugins will contribute to typegen via first class API. That means we don't need to come up with a naming convention that other plugins would be recommended to follow. We can just care about this instance.
3. Is this name too "normal looking". Should there be some indication that this is a generated package? Like `__` prefix, or `generated-` prefix, or... (I think its OK but raising in case it sparks an opinion in someone else).

Weakky: 1. Let's just reserve the package name if we agree it's a better choice (aka publish an empty package `nexus-typegen`)
3. Don't have a strong opinion here. `generated-nexus-typegen` might be a bit more explicit, but slightly more verbose/less clean 🤷‍♂

jasonkuhrt: This is an `@types` package so it would require getting into that system https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#publish-to-types.

I think we should actually do this, to help against pruning, e.g. concept of facades https://github.com/prisma/specs/issues/152.

But I think this is too high a bar for this PR. We should just move forward and track that as separate issue, motivated by resisting pruning.

> Don't have a strong opinion here. generated-nexus-typegen might be a bit more explicit, but slightly more verbose/less clean

Let's try the simple/clean for now. I think with facades we'll want this name anyways.

PR #250, opened 2019-10-06
docs: remove spectrum link

jasonkuhrt:
- consistent with what we did in nexus-prisma
- consistent with Prisma's consolidate community touch points efforts

We could add badges to the slack community?


Weakky:
Agreed on adding slack badges instead of spectrum ones 👍

jasonkuhrt:
I'm not sure how to get the slack badge up yet (left a message about that in company slack). We'll do that in a follow up PR, and it needs to be done for `nexus-prisma` too. 

As-this this at least removes link to the wrong place.

PR #268, opened 2019-10-15
fix: Default path for node_modules/@types/nexus-typegen

lewebsimple:
Simple fix inside a comment in `src/builder.ts` to indicate the proper default path for nexus-typegen.


Weakky:
Thanks for the contribution @lewebsimple, merging now 👍

PR #274, opened 2019-10-18
docs(example): show nexus + zeit + typescript

### LINKED ISSUE #270: __dirname in schema + typegen paths not settable in lambda environments
I have set up Nexus in a Zeit Now environment which I am running locally using `now dev`.

This creates a local lambda running in the /private/var folder.  Unfortunately, `__dirname` points to this folder and not my local project folder when it is executed inside of the lambda.  

This makes the generated typegen and schema inaccessible to my local project folder.  The only workaround I have found to run a separate ts-node process to generate a new typegen + schema every time I update my code.

I believe this may be the cause of https://github.com/prisma-labs/nexus-prisma/issues/382 as well.

jasonkuhrt:
Will close #270


jasonkuhrt:
CC @nikolasburk

PR #275, opened 2019-10-18
NEXUS_CONFIG env var

### LINKED ISSUE #272: Add option for exiting after Typegen step
One common question is how to generate the types without running the server. A simple option would be to add a config option to the schema:

```ts
/**
 * If set to true, or if NEXUS_EXIT env var is set to `true`,
 * we will call process.exit() after the types are emitted.
 */
exitAfterTypegen: boolean
```

And then we can tell folks to just install `ts-node` and run:

```sh
NEXUS_EXIT=true node -r ts-node/register/transpile-only path/to/schema-file.ts
```

jasonkuhrt:
This PR introduces a new nexus CLI. It is an experiment and we'll see how it goes. The first command is for handling typegen. Underscoring the experiment point, we might have [a better way to achieve this](https://github.com/prisma-labs/nexus/issues/272#issuecomment-543885816), but we can ship this value ~today, keep iterating.

will close #272

#### TODO 
- [x] fix failing-on-ci test
- [x] introduce new env var for process exit
- [x] perform process exit when configured to do so
- [x] log instructive message like we do for photon
    ```
    Generating Photon.js to /Users/jasonkuhrt/projects/prisma/nexus-prisma/examples/blog/node_modules/@generated/photon
    ```
- [x] re-write dynamic paths in test snapshots to static snappable values [e.g.](https://circleci.com/gh/prisma-labs/nexus/1670?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link):
    ```
    -     Cannot find module '/Users/jasonkuhrt/projects/prisma/nexus/entrypoint.tmp.ts'
    +     Cannot find module '/home/circleci/project/entrypoint.tmp.ts'
    ```
- [x] given nice logging from nexus on exit config, remove success logging added by cli
- [x] rename command to `generate` and add flags `-t --typescript-types` `-g --graphql-schema`. Allow each flag to override the the respective output setting. The flag argument is the output path argument.
- [x] Allow `NEXUS_CONFIG` to override `makeSchema` config
- [x] ~default file lookup system for `--entrypoint`~
- [ ] coverage


jasonkuhrt:
I tried `yargs` but was unable to get the `help` gen I wanted. I tried `commander` but was unable to get the testing I wanted. Worth mentioning both felt patchy as tools (quality etc.). Now trying oclif but now I am tripping up on https://stackoverflow.com/questions/50203369/class-constructor-cannot-be-invoked-without-new-typescript-with-commonjs. Will bump us up from es5 to es6 on the tsconfig...

bin/run.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
+
+node "%~dp0\run" %*
```
jasonkuhrt: Apparently this is for windows

jest.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
   setupFilesAfterEnv: ["<rootDir>/tests/_setup.ts"],
   preset: "ts-jest",
   testEnvironment: "node",
+  watchPathIgnorePatterns: [".*.tmp.*"],
```
jasonkuhrt: Allows the cli integration test generated files to not trigger infinite loop.

package.json
```diff
@@ -38,6 +39,7 @@
     "url": "https://github.com/tgriesser"
   },
   "dependencies": {
+    "@oclif/command": "^1.5.19",
```
jasonkuhrt: I know its another dep but I don't think it would be a good idea to roll our own here.

tests/cli/typegen.spec.ts
```diff
@@ -0,0 +1,114 @@
+import { spawnSync, SpawnSyncReturns } from "child_process";
```
jasonkuhrt: Had to resort to integration tests because oclif is tightly coupled to mocha https://oclif.io/docs/testing

tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
   "compilerOptions": {
     "downlevelIteration": true,
-    "target": "es5",
+    "target": "es6",
```
jasonkuhrt: As mentioned in pr comment to get around class error.

jasonkuhrt:
The test appears to fail on CI because of different indentation (the list expected is one column less indented for example).

jasonkuhrt:
Fixed the test by using ansi strip. Tried the env vars documented here https://github.com/chalk/chalk#chalklevel but they did not seem to work. ([oclif does use chalk v2](https://github.com/oclif/command/search?q=chalk&unscoped_q=chalk))

Not sure what's up, have created issue for better story around this in the future. https://github.com/oclif/command/issues/76.

tgriesser:
If a CLI does exist, I think I would prefer it to live outside of the Nexus library. An important goal for this project was to keep the core simple and the only hard dependencies on `graphql` (peer), `tslib`, and eventually maybe `iterall`, which is its-self a dependency of `graphql` so that'd be fine.

I don't foresee that a plugin would ever do much more in Nexus than generate the typegen, which with the addition of a config option should be no more than a one-line script. I don't think that necessitates the ceremony of `@oclif`, (as awesome of a project it is ❤️).

That's just my opinion though - possibly biased by the complaints I've received about the size of the CLI that ships with knex, that I just don't really feel light fighting that battle again if the value add isn't there.

jasonkuhrt:
@tgriesser putting PR on hold until we align on `NEXUS_CONFIG` json approach.

tgriesser:
Closing per our discussion that this sort of configuration will live outside of `nexus`, in either in the the `nexus-prisma` plugin setup or in some yet to be released CLI/framework sort of package which serves as a more opinionated layer on top of `nexus`.

Issue #188, opened 2019-07-30
Union's resolveType doesn't know about __typename


ThisIsMissEm:
I'm currently trying to build a graphql server that uses the Result Union's pattern described by [@sachee](https://twitter.com/sachee), whereby fields that may error in normal operation when resolving them return Union type of something like `union PostResult = NotFound | Post`

There's a video describing this pattern here: https://www.youtube.com/watch?v=GYBhHUGR1ZY

However, if you have:

```typescript
export const PostResult = unionType({
  name: 'PostResult',
  description:
    'A Post, or NotFound if the post was not found',
  definition(t) {
    t.members('Post', 'NotFound');
    t.resolveType(item => item.__typename);
  }
});
```

Then you get an error telling you that `__typename` isn't a property of `item`, even though your resolver for the field that uses this PostResult union can definitely return the `__typename` field.

```typescript
export const Query = objectType({
  name: 'Query',
  definition(t) {
    t.field('post', {
      type: PostResult,
      args: {
        id: idArg({ required: true })
      },
      resolve(root, { id }) {
        try {
          const post = await Posts.findById(id)
          return { __typename: "Post", ...post }
        } catch(err) {
          return { __typename: "NotFound", message: `Could not find post with ID: ${id}` }
        }
    });
  }
});
```

I think the type assigned to the callback argument in `t.resolveType` should have the `__typename` property available.

For now, I'm able to work-around this by using `item['__typename']` but that's a pretty ugly hack.

ThisIsMissEm:
Looks like it's possible to get this working by modifying `typegenHelpers.d.ts` to have:

```
export declare type ResultValue<
  TypeName extends string,
  FieldName extends string
> = GetGen3<'fieldTypes', TypeName, FieldName> & { __typename: string };
```

I've no idea how to actually get this as a proper fix though.

tgriesser:
This is indeed an oversight and something I recently also realized/encountered as well when trying to build some union types using the same pattern from that talk. 

Will look to get a fix in for this soon!

jasonkuhrt:
> I think the type assigned to the callback argument in t.resolveType should have the __typename property available.

Agree. We will need to make this invariant safe though. For example, making the resolver for a union-typed field be forced to include `__typeName` in the returned data. I expect more details to sort out than this.

jasonkuhrt:
This feels like a bug to me, more than a feature.

ThisIsMissEm:
[...]

PR #317, opened 2019-11-08
Add t.resolveType() for defaultTypeResolver

tgriesser:
Related to #188 - allows `t.resolveType()` (no args) to be used to indicate that we should use the `defaultTypeResolver` from `graphql-js` and suppress the warning.


jasonkuhrt:
Going to close this in favour of the spec laid out here: https://github.com/graphql-nexus/schema/issues/188#issuecomment-714196002

PR #324, opened 2019-11-15
feat: Add connectionPlugin for relay style pagination

tgriesser:
This plugin aims to implement the [Relay Connection Spec](https://facebook.github.io/relay/graphql/connections.htm) by default, while also being configurable both globally and on a per-field basis.

This will allow a consistent implementation of connections across a Nexus project.


examples/kitchen-sink/kitchen-sink-schema.graphql
```diff
@@ -51,16 +70,151 @@ input NestedType {
...
 }
 
+"""
+PageInfo cursor, as defined in https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
+"""
+type PageInfo {
+  """
+  The cursor corresponding to the last nodes in edges. Null if the connection is empty.
+  """
+  endCursor: String
```
swac: The [spec](https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo) says `startCursor` and `endCursor` are non-null:

> It must also contain fields startCursor and endCursor, both of which return non‐null opaque strings.

I think for this to be spec-compliant, the entire connection field on an object type needs to be nullable to account for a connection not having any edges to return.

tgriesser: > The spec says startCursor and endCursor are non-null:

I did see that, though I believe this is either incorrect or outdated. 

For instance, in graphql-relay-js, the official implementation, start and end cursor are non-null:

https://github.com/graphql/graphql-relay-js/blob/84ed0008ee17ae439ce67d820a29d1f258addff8/src/connection/connection.js#L147-L154

There is also this commit on the official relay project which demonstrates that the cursors are null when there are no results:

https://github.com/facebook/relay/commit/a17b462b3ff7355df4858a42ddda75f58c161302

There is an open PR addressing this inconsistency, but it looks like it needs an update by the person opening the PR:

https://github.com/facebook/relay/pull/2655

Thanks for taking the time to take a look though @swac, let me know if there's anything else you see in the PR that could use some improvements, I'm planning on cutting a `rc` release for it later today.

src/plugins/connectionPlugin.ts
```diff
@@ -0,0 +1,983 @@
...
+              ...fieldConfig,
+            };
+
+            let specArgs = {};
+            if (disableForwardPagination !== true) {
+              specArgs = { ...ForwardPaginateArgs };
+            }
+            if (disableBackwardPagination !== true) {
+              specArgs = { ...specArgs, ...BackwardPaginateArgs };
+            }
```
swac: Some Relay-style APIs allow making `first` a non-nullable argument if backward pagination is disabled (since the spec considers the lack of both `first` and `last` to be undefined behavior). Similarly, `last` can be non-nullable if forward pagination is disabled. Maybe this plugin can generate the types like that? Or expose a flag to enable it?

tgriesser: That's a good idea - I think a flag (`strictArgs`) that defaults to `true` (enabled) might be the best approach, since some folks may not want this behavior

For instance, in the main project I'm using Nexus on, we currently use a `perPage` / `offset` style forward-pagination. When we add this plugin, we'll want the `first` type to be loose with a custom `validateArgs` fn, so we'll want the flag to disable.

Here's the [commit](https://github.com/prisma-labs/nexus/pull/324/commits/b42c9a6455f571d95ce604ec4fe0c7dab365050e) - any reason you think it'd be better to keep it opt-in?

swac: Seems fine this way, thanks for adding it!

PR #352, opened 2019-12-19
refactor: Update warn message for inputObject

moonflare:
### What changed?
- Updated warning message when specifying `.list` and `list: true` for `inputObject`


tgriesser:
Nice, thanks!

PR #372, opened 2020-02-04
fix: wrap authorization on field subscribe

### LINKED ISSUE #358: Field Authorize Plugin executes auth after subscribing
The field authorize plugin is executing authorization on publish rather than on subscribe like on beta versions.

## Expected behavior
Authorization should fail on subscribe

## Actual behavior
Subscription fires and only fails if a publish happens

Reproduction example [here](https://github.com/AndresRodH/nexus-authorization-bug-repro)

AndresRodH:
Fix #358 

Probably we could make this configurable so consumers can opt-out of validating for subscribe -- although I don't think that'll be very common.

### DIFF
--- a/src/plugins/fieldAuthorizePlugin.ts
+++ b/src/plugins/fieldAuthorizePlugin.ts
@@ -1,4 +1,4 @@
-import { plugin } from "../plugin";
+import { plugin, CreateFieldResolverInfo, MiddlewareFn } from "../plugin";
 import {
   RootValue,
   ArgsValue,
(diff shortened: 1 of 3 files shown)


tgriesser:
Thanks for the PR!

> Probably we could make this configurable so consumers can opt-out of validating for subscribe -- although I don't think that'll be very common.

I didn't look closely, but for subscriptions, does currently only run on the `subscribe`, or on both the `subscribe` and the `resolve`? I'd imagine we'd only need/want it on the `subscribe`, right?

AndresRodH:
It would run on both. I think it'd be ideal to run it on both since there could be a scenario where a subscription could be running long enough for, for example, the user token expiring. WDYT? Would running on subscribe only make sense in this case? Is this what we could potentially configure so it's only run on either one or the other for subscriptions or both?

tgriesser:
> Currently it doesn't run on subscribe, which is what this PR addresses in order to fix the issue that's blocking us from updating nexus from beta to rc.

Gotcha. Yeah, my main question here was that I thought it was potentially confusing to have it run on both subscribe, and then subsequently on every `resolve`. Your point about validating something like a user-token is valid though, going to think about it a little more before merging.

I finally got around to implementing subscriptions for the first time ever and found quite a few things lacking, both in existing server implementations and in `graphql-js` directly [with the context](https://github.com/graphql/graphql-js/issues/894#issuecomment-584249264). I think there's some opportunity to have `nexus` simplify these concepts.

AndresRodH:
That sounds reasonable, thanks @timsuchanek. Should this PR be held off for the moment?

AndresRodH:
Looks like there's something else missing to this solutions. I decided to swap the current fieldAuthorize plugin with this PR's implementation and it looks like authorize is not running. Might have something to do with the fact that I'm not returning an asyncIterator to the subscription tests.

PR #382, opened 2020-02-18
fix: typegen always use manifest for package import

jasonkuhrt:
Follow up on 6e39cebe27397e6601d546e65e3e438734be9e5b.

### DIFF
--- a/src/typegenPrinter.ts
+++ b/src/typegenPrinter.ts
@@ -1,7 +1,9 @@
 import {
   GraphQLArgument,
+  GraphQLEnumType,
   GraphQLField,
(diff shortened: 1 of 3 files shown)


src/typegenPrinter.ts
```diff
@@ -153,13 +154,15 @@ export class TypegenPrinter {
     const importMap: Record<string, Set<string>> = {};
     const outputPath = this.typegenInfo.typegenFile;
     // For backward compat.
-    if (!this.printImports.nexus) {
+
+    const packName = getPackageNameForImport();
```
jasonkuhrt: this

Issue #392, opened 2020-03-01
connectionPlugin 'hasNextPage' returns wrong result


eyalw:
I have a an array of 6 items: [0,1,2,3,4,5] paginated using a connection field.

The user requests first: 3, and my nodes() function returns [0,1,2,3].   (1 extra, as described in the docs). and the user gets [0,1,2] as promised.
The user then requests  first: 3, after: cursor:2, I return [3,4,5] (the +1 extra is missing, indicating that there is no next page, and should return hasNextPage: false, yet nexus returns hasNextPage: true.

I'm using the latest version: 
    "nexus": "^0.12.0-rc.13",

jamiehodge:
It appears as though the default has page functions assume there is either a first or last. This is not the case if one configures them to be optional.

eyalw:
Configure who to be optional?

my code: 

```export const allAppsQuery = queryField(t => {
  t.connectionField('allApps', {
    type: 'App',
    additionalArgs: {
      status: arg({
        type: 'Boolean',
        description: 'Returns only elenents with status blah',
        required: false,
      }),
    },

    extendConnection: t => {
      t.int('totalCount', () => APPS.length);
    },

    nodes(_, { after, first }) {
      const offset = after ? Number(after) + 1 : 0;
      const results = APPS.slice(offset, offset + first + 1); // return +1 because nodes() require it
      console.log(results);

      return results;
    },

  });
});
```

jamiehodge:
I’m realizing that my previous comment was incorrect. The plugin can’t possibly know that I’ve fallen back to a slice size if there is no explicit first argument. Please ignore.

jamiehodge:
I believe the issue is the `>=` in this line ought to be `>`: https://github.com/prisma-labs/nexus/blob/develop/src/plugins/connectionPlugin.ts#L830

jamiehodge:
The documentation also refers to a non-existent option, assumeExactNodeCount: https://github.com/prisma-labs/nexus/blob/develop/src/plugins/connectionPlugin.ts#L213

tgriesser:
You should use a custom `pageInfoFromNodes` to determine the pageInfo based on how you prefer it to work. The `>=` is an intentional tradeoff for typical APIs where you're fetching the amount of rows you want given the pagination args, rather than `amount + 1` just to fulfill the `hasNextPage`.

[...]

PR #394, opened 2020-03-05
feat: base hasNextPage in connectionPlugin upon gt not gte

eyalw:
Solves #392


tgriesser:
This was an intentional tradeoff, without specifying a dedicated `pageInfoFromNodes` it's tough to know whether there are additional results without over-fetching - and the `hasNextPage` being true but not actually having any results typically isn't a big deal from a UX perspective. 

I'd say that just specifying the `pageInfoFromNodes` manually would be the best approach here.

jamiehodge:
@tgriesser your suggestion would create a fairly large amount of boilerplate for the "normal" use of this plugin, while the tradeoff is meant to support legacy API, which is arguably an edge case. Please reconsider this tradeoff. Maybe an option to support legacy API is a better option that makes the dynamics more explicit.

jasonkuhrt:
I had to make a new PR because I was unable to push test updates to this one.

PR #400, opened 2020-03-20
fix: typegen stable path generation with Windows

julien1619:
When using Windows, rootTypings paths are using `\` instead of `/` in generated code. This commit fixes it.

### DIFF
--- a/src/typegenPrinter.ts
+++ b/src/typegenPrinter.ts
@@ -170,7 +170,9 @@ export class TypegenPrinter {
         const importPath = (path.isAbsolute(val.path)
           ? relativePathTo(val.path, outputPath)
           : val.path
-        ).replace(/(\.d)?\.ts/, "");
+        )
+        .replace(/(\.d)?\.ts/, "")
+        .replace(/\\/, "/");
         importMap[importPath] = importMap[importPath] || new Set();
         importMap[importPath].add(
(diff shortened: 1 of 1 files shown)


src/typegenPrinter.ts
```diff
@@ -170,7 +170,9 @@ export class TypegenPrinter {
         const importPath = (path.isAbsolute(val.path)
           ? relativePathTo(val.path, outputPath)
           : val.path
-        ).replace(/(\.d)?\.ts/, "");
+        )
+        .replace(/(\.d)?\.ts/, "")
+        .replace(/\\/, "/");
```
jasonkuhrt: Can we use `path.normalize`?

julien1619: I tried with `normalize` first but no, because `path.normalize` is platform dependent. There is already the same fix with `replace` in this codebase.

julien1619: Here it is: https://github.com/graphql-nexus/schema/blob/develop/src/typegenAutoConfig.ts#L319

tgriesser: We should probably add `g` to the end of the regex here so it replaces all occurances: `/\\/g`

julien1619: Done.

PR #405, opened 2020-04-02
tests: add windows to os matrix

jasonkuhrt:


jasonkuhrt:
The test suites that do not pass on windows are typegenPrinter and backingTypes. I will bring them back one by test, test by test, to try and isolate and understand better the issue(s) they cause on windows.

jasonkuhrt:
Windows tests are failing for reasons outside the codebase. Found this https://github.community/t/all-jobs-failing-with-error-this-check-failed/18093.

PR #408, opened 2020-04-07
feat: allow control over nexus schema import id

jasonkuhrt:
Supports https://github.com/graphql-nexus/nexus/issues/575

Currently the typegen output is designed with the assumption that @nexus/schema will be available at node_modules/@nexus/schema. This assumes that either the project depends on @nexus/schema or that @nexus/schema has been hoisted. The former is fine but does not support being wrapped by another tool, while the latter is liable to result in module not found errors for hard-to-debug reasons (especially users who know little or nothing about the internals).


jasonkuhrt:
Looks like imports are getting sorted where they weren't before. Going to update master to make this PR diff simler.

src/plugins/connectionPlugin.ts
```diff
@@ -336,11 +336,19 @@ export const connectionPlugin = (
 
   // Define the plugin with the appropriate configuration.
 
+  console.log(
+    printedGenTypingImport({
+      module: getOwnPackage().name,
+      bindings: ["core", "connectionPluginCore"],
+    })
+  );
+
```
Weakky, line 345: I know it's wip, just a small reminder to not forget to remove that console.log 🙈

PR #416, opened 2020-04-20
feat: make `queryType` and `mutationType` merge when called several times

Weakky:
## Description

- Make `extendType` accept `description` and `nonNullDefaults`.
	- `nonNullDefaults` stays local to the `extendType` instance
    - If several `description` are passed, they joined with a `\n`
- Make `queryType` and `mutationType` use `extendType` so that when called multiple times, types are merged.

## Problems

- There's no way anymore to have one `nonNullDefaults` configuration for all `queryType` and `mutationType` extensions (as they all have their own `NonNullConfig`). The only way is to first use `objectType({ name: 'Query', nonNullDefaults: { /* ... */ } })` which kinda sucks :/

## TODO

- Figure out what broke on tests
- Looks like `extendType` of interfaces is broken, even on `develop`

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -40,6 +40,7 @@ import {
   isWrappingType,
   printSchema,
 } from "graphql";
+import { EOL } from "os";
 import {
   arg,
   ArgsRecord,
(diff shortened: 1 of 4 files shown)


Weakky:
Closing until we decide on https://github.com/graphql-nexus/nexus/issues/272#issuecomment-617076985

PR #436, opened 2020-05-10
fix(connectionPlugin): allow first and last to be zero

leonardodino:
according to the [GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm)

[4.1 Forward pagination arguments](https://relay.dev/graphql/connections.htm#sec-Forward-pagination-arguments)
> `first` takes a non‐negative integer.

[4.2 Backward pagination arguments](https://relay.dev/graphql/connections.htm#sec-Backward-pagination-arguments)
> `last` takes a non‐negative integer.

**edit**: [here's a quick example in the wild](https://graphql.org/swapi-graphql/?query=%7B%0A%20%20allFilms(first%3A%200)%7B%0A%20%20%20%20pageInfo%20%7B%0A%20%20%20%20%20%20hasNextPage%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D)

---

the patch presented here was done in a quick way to satisfy the spec,
I'm very open to feedback or improvements from the team 😄


leonardodino:
This seems to not make much sense on it's own, but it's actually really nice for counting edges with `totalCount` extension in the connection.

right now, as a workaround, I'm using `first: 1`.

Would be cool to see this merged 🙌

PR #460, opened 2020-07-02
feat: use prettier api to load config

jasonkuhrt:
This supports more prettier config scenarios. At least:

1. user is using extends feature of prettier
2. user is keeping prettier config inside package.json

### DIFF
--- a/src/typegenFormatPrettier.ts
+++ b/src/typegenFormatPrettier.ts
@@ -48,7 +48,7 @@ export function typegenFormatPrettier(
     }
 
     return prettier.format(content, {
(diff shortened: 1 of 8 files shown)


src/typegenFormatPrettier.ts
```diff
@@ -1,10 +1,13 @@
 import path from "path";
+import type * as Prettier from "prettier";
```
jasonkuhrt, line 2: This syntax required an update of prettier.

Upgrading prettier lead to some new formatting seen throughout this PR. Luckily not much.

Issue #454, opened 2020-06-19
Subscription field does not propagate payload type


Sytten:
When creating a `subscriptionField`, if one returns a properly typed `AsyncIterator` it should properly type the `root` of the `resolve` method. Unfortunately because the type is `T = any`, it seems like it overrides the real typing.

Example:
```typescript
export const MySub = subscriptionField('MySub', {
  type: SomeType,
  resolve: (payload) => {
    // payload is type object
  },
  subscribe: (_root, _args, _ctx) => {
    return createIterator() as AsyncIterator<SomeTypePayload>
  },
})
```

tgriesser:
@Sytten I believe this is fixed in 0.19 - caught some errors in the kitchen sink actually. Can you give it a shot and see if it's any better?

Sytten:
I will trust you on this, I am still on 0.14 and I don't expect I will have time to update for a while... I can test with a pet project to see though.

jasonkuhrt:
Doesn't seem to be working:

- `unknown` type for resolver param here https://github.com/graphql-nexus/nexus-plugin-prisma/blob/83c7596dc43f184562d44207af015b84b8b14cda/examples/hello-world/api.ts#L88
- and here https://github.com/graphql-nexus/nexus-plugin-prisma/blob/83c7596dc43f184562d44207af015b84b8b14cda/examples/hello-world/api.ts#L97

Sytten:
What is the return type of the method subscribe? I found that sometimes with complex/nested constructs they don't properly bubble up the type of generic parameters. I have this problem with my nexus-shield plugin on objects.

jasonkuhrt:
`AsyncGenerator<boolean, void, unknown>`

tgriesser:
@jasonkuhrt I could be wrong, but I don't know if it's possible to use the return type to infer the type usage in resolve. I could go dig up the issue, but I'm pretty sure you need to specify the type annotation of the first arg of `resolve`, it'll be type-safe (will check that the typing you supply for the `source` is correct), it's just not inferred.

jasonkuhrt:
I think you are right!

Sytten:
I thought that TS was able to infer a generic parameter from the output but it doesn't seem like it. :(

tgriesser:
Looks like a TS bug actually https://github.com/microsoft/TypeScript/issues/41456

tgriesser:
[...]

Issue #78, opened 2019-03-09
SubscribeFieldConfig resolve shouldn't be required


richardjli:
Thanks for implementing subscriptions -- these are the missing piece for an app I'm working on.

However, I'm a little confused by why the resolve property on SubscribeFieldConfig is required. Resolve in subscription resolvers is mostly useless, everything is handled through subscribe. Is there something I'm missing about how this is supposed to be used?

tgriesser:
> Resolve in subscription resolvers is mostly useless, everything is handled through subscribe.

As I understand it, this isn't always true, `subscribe` sets up the subscription, but the `resolve` is used to format the data from the published event into the actual shape needed to fulfill the subscription query. If these do map 1:1, for now you can just add an identity function: `resolve: t => t` as a property the config.

I don't have any strong opinions here, if this is a common enough case I can consider making this optional - didn't really spend as much time thinking about the guarantees around type-safety in subscriptions as in the rest of the schema.

Devlin556:
@tgriesser Yesterday, I also stuck with this problem. My subscription returns data from the async iterator as it's needed for the target type. Maybe resolver should be `t => t` by default?

jasonkuhrt:
Ideally `resolve` is required when the event payloads returned by subscribe are not 1:1, and otherwise optional (possibly even hidden).

Achieving this will require some advanced techniques though, and would probably be out of scope for `@nexus/schema`.

Need to think it through but roughly I think: 1) we can use the TSC API to extract return type from `subscribe` 2) compare with return type of what resolve wants 3) require resolve if they differ meaningfully.

That is non-trivial to make robust and performant, but not so bad.

We'd want `@nexus/schema` to make this possible but the work involved here is, again, out of scope. Only Nexus framework goes to the level of using the TSC API right now.

I think @tgriesser's suggestion is a great one for `@nexus/schema` users.

I'll leave this issue here for a bit, but probably migrate it to the Nexus repo soon.

PR #462, opened 2020-07-06
feat: add subscriptionType

jasonkuhrt:
supports https://github.com/graphql-nexus/nexus/issues/447

### About the implementation

- graphql.js treats subscription type as an object type
- at least, for example, `GraphQLFieldConfig` type in graphql has optional `subscribe` field

	image

- for this reason, in builder, we share the `buildObjectType` in Nexus, and it just works

	Screenshot of playground for local sandbox app using linked `@nexus/schema` and working

	image

- I briefly considered reusing the existing `NexusObjectTypeDef` and genericizing its field config type –– but I found this to add more complexity and type errors across the codebase as I began in that direction.

	For example, the union of being either subscription config or object config made `implements` method need rethinking.

	image

### Note

I do not think the story about Nexus & subscription support is done with this PR.

For example the typings on the resolve method are `any` which is wrong––the `root` param on resolvers of subscription root fields should be typed to the return value of the `subscribe` field.

And that is already an existing issue with [`subscriptionField`](https://github.com/graphql-nexus/schema/issues/454).

I strongly believe we should iterate instead of letting this PR live for long.


src/definitions/subscriptionType.ts
```diff
@@ -0,0 +1,77 @@
...
+   */
+  resolve(
+    root: T,
+    args: ArgsValue<TypeName, FieldName>,
+    context: GetGen<'context'>,
+    info: GraphQLResolveInfo
+  ):
+    | MaybePromise<ResultValue<'Subscription', FieldName>>
+    | MaybePromiseDeep<ResultValue<'Subscription', FieldName>>
+}
```
Weakky, line 31: I know it's not really part of the pr goal, but can we maybe make `resolve` optional and provide an "identity" default value?

Weakky: Saw your comment on the PR. Ok to iterate further down the road

src/builder.ts
```diff
@@ -286,6 +289,7 @@ export type TypeToWalk =
   | { type: 'named'; value: GraphQLNamedType }
   | { type: 'input'; value: NexusShapedInput }
   | { type: 'object'; value: NexusShapedOutput }
+  | { type: 'subscription'; value: NexusSubscriptionTypeConfig }
```
Weakky, line 292: Given that subscription is just an object type, what's the reason for needing an additional type to walk here? It looks like it's not even walking the type in a special way. Can't we just type the surface API but then return a simple `ObjectTypeDef` ?

src/definitions/subscriptionType.ts
jasonkuhrt: Funny you say that --> https://github.com/graphql-nexus/schema/issues/78#issuecomment-654005333

src/definitions/subscriptionType.ts
```diff
@@ -74,9 +74,7 @@ export interface SubscribeFieldConfig<TypeName extends string, FieldName extends
 
 export interface SubscriptionDefinitionBuilder extends ObjectDefinitionBuilder<'Subscription'> {}
 
-export class SubscriptionDefinitionBlock<
-  TypeName extends string = 'Subscription'
-> extends ObjectDefinitionBlock<TypeName> {
+export class SubscriptionDefinitionBlock extends ObjectDefinitionBlock<'Subscription'> {
```
jasonkuhrt, line 77: @Weakky can you explain why this fix worked?

PR #473, opened 2020-08-02
feat: allow asNexusMethod to specify TS type

### LINKED ISSUE #469: Improve typing of custom scalars
Related to https://github.com/graphql-nexus/schema/issues/360

Currently the typing of custom scalars is not great. But it is already possible to pass `rootTyping` to the `scalarType` which adds the type to the `NexusGenRootTypes` but then it does not use it when we actually use the type. We can use the parameter `backingTypeMap`, but it feels unnecessary since we already provide the root typing.

So my proposition is two fold:
- Use the `rootTyping` when available 
- Accept a root typing in the method `asNexusMethod`

Sytten:
fixes #469 

## Goal
- Improve the general typing of custom scalars

## Changes
- Create a new `NexusGenScalars` for all scalar types
- Remove scalars from `NexusGenRootTypes` (might be a breaking change, need review)
- Use the `NexusGenScalars` for all non specified (base GraphQL) scalars in `input` and `output` types
- Add `rootTyping` parameter to `asNexusMethod` to allow the user to specify it, otherwise it will fallback to the `backingTypeMap`

BREAKING CHANGE:

The global TS type `NexusGenRootTypes` no longer contains scalars. All scalars now live under a new global TS type named `NexusGenScalars`.

```ts
// before
let foo: NexusGenRootTypes['String']
```
```ts
// after
let foo: NexusGenScalars['String']
```

### DIFF
--- a/src/definitions/scalarType.ts
+++ b/src/definitions/scalarType.ts
@@ -52,8 +52,13 @@ export function scalarType<TypeName extends string>(options: NexusScalarTypeConf
   return new NexusScalarTypeDef(options.name, options)
 }
 
-export function asNexusMethod<T extends GraphQLScalarType>(scalar: T, methodName: string): T {
+export function asNexusMethod<T extends GraphQLScalarType>(
+  scalar: T,
+  methodName: string,
+  rootTyping?: RootTypingDef
+): T {
   return decorateType(scalar, {
     asNexusMethod: methodName,
+    rootTyping: rootTyping,
   })
 }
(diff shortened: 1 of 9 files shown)


Weakky:
Content looks good to me. It's def. a breaking change if anyone was referencing a scalar type from `core.GetGen<'rootTypes'>`. What I'm not 100% sure right now though is if there's any part of the internal typings that are expecting the scalar types to be in the `rootTypes`. Maybe @tgriesser has better insights here?

jasonkuhrt:
> AFAIK, the root arg of resolvers is typed using the rootTypes, but it doesn't make sense to me that a scalar would ever appear as a root/parent value in a resolver, which makes me think the PR is fine!

Agreed!

LGTM

I think we should merge this and get it into today's release which already includes one breaking change.

PR #496, opened 2020-08-27
feat: allow interfaces to implement other interfaces

santialbo:
This is a initial draft for allowing interfaces to implement other interfaces which is one of the features coming with graphql v15.
Things I've considered:
- No need to call `t.implements` with all the interfaces like it is required on grapqhl. It automatically pulls all the transitive "implements" when building the schema.
- Check for circular dependencies before building the types in order to avoid stack overflow exceptions.

Please let me know if you prefer this to be handled in another way or if you would like me to add more tests and what kind of tests you would like to see.


tgriesser:
Cool stuff, definitely want to get this in here. Mind getting it up-to-date with develop?

> No need to call t.implements with all the interfaces like it is required on grapqhl. It automatically pulls all the transitive "implements" when building the schema.

Sounds like a good feature!

> Check for circular dependencies before building the types in order to avoid stack overflow exceptions.

Nice

> BTW: I missed having some sort of autoformatter when working on this. Seeing that you use prettier on graphql-nexus/nexus would you accept a PR adding it here with the same config?

Yeah, it definitely used to be in here. I'm guessing with all the shuffling around with Prisma things it got lost in the mix. Feel free to add!

jasonkuhrt:
Tests appear to be passing. NPM is having issues right now so can ignore that failure:

image

jasonkuhrt (approved):
Thanks @santialbo !

PR #508, opened 2020-09-22
feat: output types & list items are now nullable by default 

Weakky:
This PR

- makes output types nullable by default
- makes list items nullable by default

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -420,7 +420,7 @@ export class SchemaBuilder {
   constructor(protected config: BuilderConfig) {
     this.nonNullDefaults = {
       input: false,
-      output: true,
+      output: false,
       ...config.nonNullDefaults,
     }
     this.plugins = config.plugins || [fieldAuthorizePlugin()]
@@ -1147,7 +1147,7 @@ export class SchemaBuilder {
   protected decorateList<T extends GraphQLOutputType | GraphQLInputType>(type: T, list: true | boolean[]): T {
     let finalType = type
     if (!Array.isArray(list)) {
-      return GraphQLList(GraphQLNonNull(type)) as T
+      return GraphQLList(type) as T
     }
     if (Array.isArray(list)) {
       for (let i = 0; i < list.length; i++) {
--- a/tests/plugins/nullabilityGuardPlugin.spec.ts
+++ b/tests/plugins/nullabilityGuardPlugin.spec.ts
@@ -92,12 +92,12 @@ const types = [
   }),
   queryField('intList', {
     type: 'Int',
-    list: true,
+    list: [true],
     resolve: () => [1, 2, null],
   }),
   queryField('userList', {
     type: 'User',
-    list: true,
+    list: [true],
     resolve: () => [null, Promise.resolve(null), null],
   }),
   queryField('interfaceType', {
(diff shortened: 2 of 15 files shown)


src/builder.ts
```diff
@@ -1147,7 +1147,7 @@ export class SchemaBuilder {
   protected decorateList<T extends GraphQLOutputType | GraphQLInputType>(type: T, list: true | boolean[]): T {
     let finalType = type
     if (!Array.isArray(list)) {
-      return GraphQLList(GraphQLNonNull(type)) as T
```
Weakky, line 1150: bug fix for making list items non-nullable by default

src/builder.ts
```diff
@@ -420,7 +420,7 @@ export class SchemaBuilder {
   constructor(protected config: BuilderConfig) {
     this.nonNullDefaults = {
       input: false,
-      output: true,
+      output: false,
```
Weakky, line 423: makes output types nullable by default

jasonkuhrt (approved):
LGTM 

Would be nice if @tgriesser can have a look too

We should merge this week I think.

tests/plugins/nullabilityGuardPlugin.spec.ts
```diff
@@ -92,12 +92,12 @@ const types = [
...
   queryField('intList', {
     type: 'Int',
-    list: true,
+    list: [true],
     resolve: () => [1, 2, null],
   }),
   queryField('userList', {
     type: 'User',
-    list: true,
+    list: [true],
```
tgriesser, line 100: The array syntax for `list` has always felt kind of awkward. Was thinking we could deprecate the `list` property on the object here, instead adding `list` and `nonNull` helper fns similar to graphene https://docs.graphene-python.org/en/latest/types/list-and-nonnull/, so this would become:

```ts
import { list, queryField } from '@nexus/schema'

queryField('userList', {
  type: list('User'),
  resolve: () => [null, Promise.resolve(null), null],
})
```

Haven't taken a shot at it yet, but I don't think it'd be too difficult to type / unwrap as necessary.

tgriesser (changes requested):

tests/plugins/nullabilityGuardPlugin.spec.ts
jasonkuhrt: Could we defer this change to another issue/PR. Agree that we have room to improve here. But I don't think this is necessary to tackle here or in order to ship nullability defaults change.

@tgriesser OK with you?

homoky: > The array syntax for `list` has always felt kind of awkward. Was thinking we could deprecate the `list` property on the object here, instead adding `list` and `nonNull` helper fns similar to graphene https://docs.graphene-python.org/en/latest/types/list-and-nonnull/, so this would become:
> 
> ```ts
> import { list, queryField } from '@nexus/schema'
> 
> queryField('userList', {
>   type: list('User'),
>   resolve: () => [null, Promise.resolve(null), null],
> })
> ```
> 
> Haven't taken a shot at it yet, but I don't think it'd be too difficult to type / unwrap as necessary.

Why to drop it? Is it hard to keep those two options?

tgriesser: @homoky we can keep it around for a bit (or forever) I suppose - just like the idea of consolidating on one way to do things when it's practical.

PR #533, opened 2020-10-03
feat: add onObjectDefinition / onInputObjectDefinition

tgriesser:
The `onObjectDefinition` plugin option allows for a plugin to declarative modify the definition of an object type.

For example:

```ts
const Post = objectType({
  name: 'Post',
  node: 'uuid',
  resolve(t) {
    t.string('name')
  }
})
```

This allows the plugin to intercept the `t` (definition block builder) after the fields are applied, and based on the presence of `node` in the object definition, add: 

```ts
onObjectDefinition((t, config) => {
  if (config.node) {
    t.implements('Node')
    t.id('id', { 
      nullable: false,
      resolve: (root, args, ctx, info) => `${info.parentType.name}:${root[config.node]}`
    })
  }
})
```

Also added `onInputObjectDefinition` for symmetry, but not sure if it'd be useful yet.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,6 +20,7 @@ jobs:
         run: yarn -s test:ci
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
+        if: matrix.os == 'ubuntu-latest' && matrix.node-version == '10.x'
         with:
           directory: ./coverage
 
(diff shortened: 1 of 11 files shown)


jasonkuhrt:
Awesome @tgriesser! Will look at this tomorrow.

tgriesser:
@jasonkuhrt thinking about it more, it seems like this sort of thing could also be useful for Prisma, especially if we added a way to determine what fields may have been already defined in the "definition" block.

```ts
export const User = objectType({
  prismaModel: 'User',
})

export const Post = objectType({
  prismaModel: 'Post',
  definition(t) {
    t.string('title', {
       resolve: customResolver
    })
  }
})
```

PR #535, opened 2020-10-04
chore(docs): Fix a prisma client call b/c latest Prisma Client

cybaj:
Update post query due to latest Prisma Client which of version is `2.8.0`.


jasonkuhrt:
Thanks!

PR #538, opened 2020-10-05
feat: replace list & nullable API by type wrappers

### LINKED ISSUE #591: Improve nullability & list API
## Motivation

Current API tries to describe nullability & list using booleans.

```ts
t.field('foo', {
  nullable: boolean,
  list: boolean | boolean[]
})
```

Here's how the current API can be used:

```ts
import { objectType } from '@nexus/schema'

export const Test = objectType({
  name: 'Foo',
  definition(t) {
    t.string('a') // String
    t.string('b', { nullable: false }) // String!
    t.string('c', { list: true }) // [String]
    t.string('d', { list: true, nullable: false }) // [String]!
    t.string('e', { list: [true] }) // [String!]
    t.string('f', { nullable: false, list: [true] }) // [String!]!
  }
})
```

This can quickly become very unintuitive. I actually had to quickly jump in a playground to remember the API.

## Proposal n°1

Replace the current API with the following two simple constructors: `nonNull()` & `list()`, inspired from https://docs.graphene-python.org/en/latest/types/list-and-nonnull/.

While its usage is slightly more verbose, it's also much clearer.

**Usage example**:

```ts
import { objectType, list, nonNull } from '@nexus/schema'

objectType({
  name: 'Foo',
  definition(t) {
    t.field('a', { type: 'String' }) // String
    t.field('b', { type: nonNull('String') }) // String!
    t.field('c', { type: list('String') }) // [String]
    t.field('d', { type: list(nonNull('String')) }) // [String!]
    t.field('e', { type: nonNull(list('String')) }) // [String]!
    t.field('f', { type: nonNull(list(nonNull('String'))) }) // [String!]!
  }
})
```

The biggest con with this proposal is that it doesn't seem to support the `t.{scalars}` shorthands at all since the types are now wrapped in `list()` and `nonNull()` types. Here's a quick illustration:

```ts
t.string('myField', { /*no type property, so we can't express list or nullability*/ })
```

Did I maybe miss something that would keep this proposal usable with the scalar shorthand? We _could_ enable something like:

```ts
t.string('somefield', {
  listOrNullability: list(nonNull()) // [String!] - Please, ignore the terrible `listOrNullability` field name, it doesn't matter here
})
```

But I'm not sure that's intuitive either...

## Proposal n°2

The current API already supports shorthand for list fields, using the `t.list.*`. We could expand on the previous idea and allow the following expressions:

```ts
import { objectType } from '@nexus/schema'

objectType({
  name: 'Foo',
  definition(t) {
    t.string('a') // String
    t.nonNull.string('b') // String!
    t.list.string('c') // [String]
    t.list.nonNull.string('d') // [String!]
    t.nonNull.list.string('e') // [String]!
    t.nonNull.list.nonNull.string('f') // [String!]!
  },
})
```

I honestly don't like very much the chaining style of this API, which makes the readability fairly hard. But at least it keeps supporting the scalar shorthands which I find very useful to reduce the boilerplate already created by the code-first paradigm.

## Proposal n°3

It's very unlikely that we'll want that, but just to be exhaustive, we could as well allow the SDL representations:

```ts
import { objectType, list, nonNull } from '@nexus/schema'

objectType({
  name: 'Foo',
  definition(t) {
    t.field('a', { type: 'String' }) // String
    t.field('b', { type: 'String!' }) // String!
    t.field('c', { type: '[String]' }) // [String]
    t.field('d', { type: '[String!]' }) // [String!]
    t.field('e', { type: '[String]!' }) // [String]!
    t.field('f', { type: '[String!]!' }) // [String!]!
  }
})
```

Cons:

- No support for scalar shorthands either
- No support for type reference using the actual values (as opposed to references using type names)
- Could blow up the union type member threshold very quick (which is around 100k I think)
- Cannot type-safely support nested arrays (which could be infinitely nested) without blowing up the union type member threshold even quicker

Let us know if you have better proposals and/or which one you prefer here 🙌

Weakky:
closes #591

## TODO

- [x] Figure out how to deal with `nonNullDefaults`
- [x] Prevent `nonNull(nonNull())`
- [x] add `nullable()`
- [x] support args
- [x] update tests
- [x] update examples
- [x] update docs
- [ ] update SDL converter
- [ ] mark squashed commit as BREAKING

## Notes

- `t.connectionField` keeps the `nullable: boolean` API
- We currently don't type-check for what's inside a wrapped type. That means this is currently possible at build-time: `t.field({ type: list(stringArg()) })`


src/definitions/list.ts
```diff
@@ -0,0 +1,60 @@
...
+
+type ListableTypes = AllNamedTypes | NexusNonNullDef<NonNullableTypes> | NexusListDef<ListableTypes>
+
+export type NexusListDefConfig<TypeName extends ListableTypes> = {
+  type: TypeName
+}
+
+export class NexusListDef<TypeName extends ListableTypes> {
+  // @ts-ignore
+  private _isNexusListDef: boolean = true
```
Weakky, line 19: Needed to differentiate it from the `NexusNonNullDef`, otherwise, as both classes have the same constructor, TS doesn't differentiate them

src/definitions/list.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function list<TypeName extends ListableTypes>(type: TypeName) {
+  return new NexusListDef(type)
+}
+
+/**
+ * nonNull()
+ */
+
+type NonNullableTypes = AllNamedTypes | NexusListDef<ListableTypes>
```
tgriesser, line 34: I wonder if we should just add `NexusNonNullDef<NonNullableTypes>` to this, so you can wrap `nonNull(nonNull(...))` and have it be a no-op. IIRC that's how GraphQL's definition of non-nullable works

Weakky: Ohhh, I didn't know about that! Yeah, that would also simplify the types 👍

tgriesser: Turns out I was completely wrong about this. Don't know what I thought I was remembering. Thanks to @Weakky for pointing this out 🤦

src/utils.ts
```diff
@@ -505,6 +527,155 @@ export function isNodeModule(path: string) {
...
+  let namedType = argDef
+
+  while (isNexusWrappingType(namedType)) {
+    namedType = namedType.ofType
+  }
+
+  return namedType
+}
+
+export function getNexusNamedType(type: AllNexusTypeDefs | string): AllNexusNamedTypeDefs | string {
```
jasonkuhrt, line 665:
```suggestion
/**
 * Strip away list null and nonNull types recursively until another type like an enum or object  
(so-called "named types") is found, then return.
 */
export function unwrapUntilFirstNamedOutputType(type: AllNexusTypeDefs | string): AllNexusNamedTypeDefs | string {
```

src/utils.ts
```diff
@@ -505,6 +527,155 @@ export function isNodeModule(path: string) {
...
+
+    if (kind.type === 'List' && kind.nullable === false) {
+      return nonNull(list(type))
+    }
+
+    return type
+  }, baseType)
+}
+
+export function getNexusNamedArgDef(argDef: AllNexusArgsDefs | AllInputTypes): AllNexusNamedArgsDefs {
```
jasonkuhrt, line 651: `unwrapUntilFirstNamedInputType`?

src/utils.ts
```diff
@@ -505,6 +527,155 @@ export function isNodeModule(path: string) {
...
+    // Type
+    if (!isNexusWrappingType(namedType)) {
+      wrapping.unshift({ type: 'WrappedType', nullable: !nonNullDefault })
+    }
+  }
+
+  return { namedType, wrapping }
+}
+
+export function wrapType(
```
jasonkuhrt, line 588:
```suggestion
/**
 * Take a type and adjust the wrapping to correctly reflect the nullability defaults. For example a list(Foo) with nullable by default would become nullable(list(Foo)).
 */
export function rewrapTypeWithNullabilityDefaults(
```

src/utils.ts
```diff
@@ -505,6 +527,155 @@ export function isNodeModule(path: string) {
...
+type NexusWrapKind = { type: 'List'; nullable: boolean } | { type: 'WrappedType'; nullable: boolean }
+
+export function unwrapNexusDef(
+  typeDef: AllNexusTypeDefs | AllNexusArgsDefs | string,
+  nonNullDefault: boolean
+): { namedType: AllNexusNamedTypeDefs | AllNexusArgsDefs | string; wrapping: NexusWrapKind[] } {
+  const wrapping: NexusWrapKind[] = []
+  let namedType = typeDef
+
+  const isWrappedType = (obj: any): obj is AllNexusNamedTypeDefs | NexusArgDef<any> | string =>
```
jasonkuhrt, line 539:
```suggestion
  const isNamedType = (obj: any): obj is AllNexusNamedTypeDefs | NexusArgDef<any> | string =>
```

jasonkuhrt: Or negate meaning here and then down below do `!isWrappedType(...)` which seems easier to understand.

src/definitions/definitionBlocks.ts
```diff
@@ -173,25 +159,20 @@ export class OutputDefinitionBlock<TypeName extends string> {
 
   protected decorateField(config: NexusOutputFieldDef): NexusOutputFieldDef {
     if (this.isList) {
-      if (config.list) {
+      if (isNexusListTypeDef(config.type)) {
```
jasonkuhrt, line 162: Should we be qualifying our internal function names like `isNexus...`?

jasonkuhrt (approved):
LGTM

I didn't go deep on the builder changes.

Great work!

docs/content/index.mdx
```diff
@@ -82,12 +83,4 @@ const Query = queryType({
...
 })
 
-// Recursively traverses the value passed to types looking for
-// any valid Nexus or graphql-js objects to add to the schema,
-// so you can be pretty flexible with how you import types here.
-const schema = makeSchema({
-  types: [Account, Node, Query, StatusEnum],
-  // or types: { Account, Node, Query }
-  // or types: [Account, [Node], { Query }]
-})
```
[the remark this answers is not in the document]
Weakky: That was an accident, I added it back already 👍

src/builder.ts
```diff
@@ -1108,13 +1124,10 @@ export class SchemaBuilder {
...
     }
     const fieldExtension = new NexusFieldExtension(fieldConfig)
+    const nonNullDefault = this.getNonNullDefault(typeConfig, 'output')
     const builderFieldConfig: Omit<NexusGraphQLFieldConfig, 'resolve' | 'subscribe'> = {
       name: fieldConfig.name,
-      type: this.decorateType(
-        this.getOutputType(fieldConfig.type),
-        fieldConfig.list,
-        this.outputNonNull(typeConfig, fieldConfig)
-      ),
```
tgriesser, line 1117: Mentioned in Slack, but similar question to the 2nd arg as a function removal... do we want to detect the presence of `list / nullable` in the existing object and keep the existing logic... making it a type error but a runtime warning for a bit (with some `TODO: remove` in the code so we remember to remove it)?

tgriesser (changes requested):
Looks solid!

src/builder.ts
```diff
@@ -1099,6 +1103,18 @@ export class SchemaBuilder {
...
+  protected getNonNullDefault(
+    typeConfig:
+      | NexusGraphQLObjectTypeConfig
+      | NexusGraphQLInterfaceTypeConfig
+      | NexusGraphQLInputObjectTypeConfig,
+    kind: 'input' | 'output'
+  ): boolean {
+    const { nonNullDefaults = {} } = typeConfig.extensions?.nexus?.config || {}
+
+    return firstDefined(nonNullDefaults[kind], this.config.nonNullDefaults[kind], false)
```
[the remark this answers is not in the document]
Weakky: Yeah, this is mostly using some helpers that @tgriesser was using. Had to replicate that logic in the connectionPlugin and ended up using the `??` operator there. Will refactor

src/builder.ts
```diff
@@ -1099,6 +1103,18 @@ export class SchemaBuilder {
...
   }
 
+  protected getNonNullDefault(
+    typeConfig:
+      | NexusGraphQLObjectTypeConfig
+      | NexusGraphQLInterfaceTypeConfig
+      | NexusGraphQLInputObjectTypeConfig,
+    kind: 'input' | 'output'
+  ): boolean {
+    const { nonNullDefaults = {} } = typeConfig.extensions?.nexus?.config || {}
```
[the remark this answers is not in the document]
Weakky: Could you elaborate what you don't understand? 🤔

src/utils.ts
Weakky: This named after the `getNamedType()` function from `graphql-js` which achieves the same thing. It's a super commonly used function in all GraphQL tools so I'd personally prefer keeping it named that way

src/utils.ts
Weakky: Same as above

src/definitions/definitionBlocks.ts
Weakky: I think this is to differentiate from the `graphql-js` type defs

examples/ghost/src/schema/User.ts
```diff
@@ -1,35 +1,35 @@
...
-    t.string("locale", { nullable: true });
-    t.string("visibility");
-    t.string("metaTitle", { nullable: true });
-    t.string("metaDescription", { nullable: true });
-    t.string("tour", { nullable: true });
-    t.date("lastSeen", { nullable: true });
-    t.date("createdAt");
-    t.field("createdBy", {
-      type: User,
+    t.field('id', { type: nonNull('String') })
```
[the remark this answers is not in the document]
jasonkuhrt: Yep remember our convo now about the two PRs in one canary etc.

src/utils.ts
jasonkuhrt: Oh I see, fair enough yeah.

tests/typegenPrinter.spec.ts
```diff
@@ -19,7 +20,12 @@ describe('typegenPrinter', () => {
           __typename: true,
         },
       },
-      prettierConfig: path.join(__dirname, '../.prettierrc'),
+      async formatTypegen(source, type) {
```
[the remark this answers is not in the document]
Weakky: This is a fix after merging the abstract type PR. Basically, since we added the `Discriminate<>` type helper, we `import { core } from @nexus/schema` when types are needed to be discriminated, which is the case on that test. So before, we weren't importing anything from @nexus/schema but now we do.

I know it's not related to that PR but it was creating some type errors in some places

src/builder.ts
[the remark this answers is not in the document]
jasonkuhrt: All good, maybe on a call sometime

examples/ghost/src/generated/ghost-nexus.ts
```diff
@@ -58,43 +58,43 @@ export type NexusGenAllTypes = NexusGenRootTypes & NexusGenScalars
...
+    userById: NexusGenRootTypes['User'] // User!
   }
   User: {
     // field return type
     accessibility: string | null // String
     bio: string | null // String
     coverImage: string | null // String
-    createdAt: NexusGenScalars['Date'] | null // Date
-    createdBy: NexusGenRootTypes['User'] | null // User
-    email: string | null // String
```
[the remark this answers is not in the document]
Weakky: Yes and no. This example, like others, had their typegen file changed when we switch the nonNullDefaults to be false by default. I made this example use nonNullDefaults: { output: true } again, so that the usages of nullable() make sense (just like we were doing nullable: true before).

After making output types nullable by defaults, there were many places where we were doing nullable: true although the field was already nullable. I did not create these examples but wanted to reflect again what was initially meant to be done

src/utils.ts
Weakky: > Take a type and adjust the wrapping to correctly reflect the nullability defaults. For example a list(Foo) with nullable by default would become nullable(list(Foo))

This function is not only there to reflect the nullability defaults. It's mainly here to convert nexus wrappers to GraphQL wrappers (and while doing that, we also reflect the nullability defaults indeed)

Weakky: I suggest renaming it to `wrapAsGraphQLType`, which mirrors the existing `wrapAsNexusType`

Weakky: or `rewrapAsGraphQLType`

Weakky: named to `rewrapAsGraphQLType`

src/utils.ts
Weakky: done!

Weakky: The reason it was called `WrappedType` is because there can be NexusArgDef in here, which technically aren't named types. But it doesn't matter

src/builder.ts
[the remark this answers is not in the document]
Weakky: done!👍

docs/content/015-api/061-list-nonNull-nullable.mdx
```diff
@@ -0,0 +1,127 @@
+---
+title: list / nonNull / nullable
+codeStyle: true
+---
+
+## list
+
+[GraphQL Docs on List Type](https://graphql.org/learn/schema/#lists-and-non-null)
+
+Types and Arguments can be passed to the `list()` helper to mark them as list.
```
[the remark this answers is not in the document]
Weakky: done 👍

docs/content/015-api/061-list-nonNull-nullable.mdx
```diff
@@ -0,0 +1,127 @@
...
+    })
+  }
+})
+```
+
+⚠️ If `nonNullDefaults.input` or `nonNullDefaults.output` is true, `list('String')` will resolve to `nonNull(list(nonNull('String')))`, producing the following type: `[String!]!`
+
+Below are some usage examples of `list()`:
+
+| Examples                           | Produced GraphQL Type | nonNullDefaults |
```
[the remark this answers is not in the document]
Weakky: done 👍

src/builder.ts
jasonkuhrt: Does this matter less if we're aiming for a 1.0 cut on December 11?

But I agree this would be nice though

PR #557, opened 2020-10-15
fix: error if root typing path is invalid

### LINKED ISSUE #396: A wrong `rootTyping.path` does not report any error/warning
### Description

Consider the following code:

```ts
// src/graphql.ts
import { objectType } from 'nexus'

export type CustomUser = { /* ... */ }

objectType({
  name: 'User',
  rootTyping: {
    path: __dirname,
    name: 'CustomUser'
  },
  definition(t) { ... }
})
```

The use of `__dirname` will generate a wrong import in the typegen file. It would only work if the file was an `index.ts` file. `__filename` should be used instead.

However, this create some TS errors that doesn't help at all and no warning is provided by the runtime either.

**TS error**

image

### Proposal

If `rootTyping.path` doesn't resolve to a file, a warning should be logged to help users fix their problem.

Weakky:
Closes #396

We unfortunately cannot add a runtime error if the type that's imported from the path does not exist without substantial work and/or performance overhead.

If we use the current regex approach, we might end up throwing in cases where the type is properly exported but the regex fails to see that.

The only approach would be to parse the actual file with the type-checker, which is way out of scope for this PR.

### DIFF
--- a/src/typegenPrinter.ts
+++ b/src/typegenPrinter.ts
@@ -1,3 +1,4 @@
+import fs from 'fs'
 import {
   GraphQLArgument,
   GraphQLEnumType,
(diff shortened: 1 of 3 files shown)


src/typegenPrinter.ts
```diff
@@ -159,13 +160,26 @@ export class TypegenPrinter {
...
+          throw new Error(`Root typing path ${rootTypePath} of the type ${typeName} does not exist`)
+        }
+
+        const importPath = relativePathTo(rootType.path, outputPath)
           .replace(/(\.d)?\.ts/, '')
           .replace(/\\+/g, '/')
+
         importMap[importPath] = importMap[importPath] || new Set()
-        importMap[importPath].add(val.alias ? `${val.name} as ${val.alias}` : val.name)
+        importMap[importPath].add(rootType.alias ? `${rootType.name} as ${rootType.alias}` : rootType.name)
```
Weakky, line 182: This is technically a breaking change as I'm now only accepting absolute file paths, to ensure the ability to check whether the file exists.

@tgriesser I'm not sure what was the use-case for enabling any kind of path. Path aliases maybe? Do you feel strongly about keeping any string allowed? In that case, I'll make the checks only in case the path is absolute

jasonkuhrt (approved):
LGTM, let's wait a bit to see if @tgriesser has thoughts too.

jasonkuhrt:
@Weakky since there were no tests docs or comments to indicate otherwise I think we can proceed with this PR.

Weakky:
@jasonkuhrt I was planning to mention the breaking change in the squashed commit

jasonkuhrt:
@Weakky I don't think we have to consider them breaking because it was never documented or tested for that use-case. But you're right that it would be nice to mention. Let's add some detail in the release notes when we cut stable?

PR #563, opened 2020-10-19
fix: input fields with default should not be typed as optional

### LINKED ISSUE #127: Default Arg Value Type Error
Given the following objectType:

```ts
import { extendType, intArg, stringArg } from 'nexus'
import { Research } from '../db/entities/comptool/Research'

export const FilterQueries = extendType({
  type: 'Query',
  definition(t) {
    t.list.string('companies', {
      description:
        'Default returns the first 20 in alphabetical order. Searchable',
      args: {
        limit: intArg({
          default: 20,
          required: false,
        }),
        offset: intArg({
          default: 0,
          required: false,
        }),
        search: stringArg({
          default: '',
          required: false,
        }),
      },
      resolve: async (_root, { limit, offset, search }) => {
        return Research.findCompaniesFilter(limit, offset, search).then(
          results => {
            return results.map(r => r.current_company)
          },
        )
      },
    })
  },
})
```

Why would `limit,` `offset,` `search` have a type of `string(or number) | null | undefined` when there's a default value?

Weakky:
Closes #127

### DIFF
--- a/src/typegenPrinter.ts
+++ b/src/typegenPrinter.ts
@@ -405,7 +405,7 @@ export class TypegenPrinter {
             if (!this.hasResolver(field, type)) {
               if (typeof obj !== 'string') {
                 obj[field.name] = [
-                  this.argSeparator(field.type as GraphQLInputType),
+                  this.argSeparator(field.type as GraphQLInputType, false),
                   this.printOutputType(field.type),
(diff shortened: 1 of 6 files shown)


jasonkuhrt (approved):
LGTM

PR #564, opened 2020-10-19
fix: subscription type static typings

### LINKED ISSUE #559: subscription block t-helpers missing "subscribe" field typing
For example, in the following, `t.field` works but `t.boolean` does not––it "thinks" `subscribe` is an invalid field.

image

jasonkuhrt:
fixes #559

This takes another stab at implementing subscription type. It fixes multiple things.

- It no longer sub-classes from Object Type. This means we lose interface implements support. However this was never tested before and it is unclear its even a feature we want.
- It simplifies the field shorthand typings by removing the resolver shorthand param style. This style is unusable since subscribe config field is always needed anyways.
- It fixes field shorthand config which would fail when typegen was present. Specifically the subscribe field was not being recognized before.

Also

- It adds subscription type test cases to integration test. This way we capture the error that only happens when typegen is present.

### DIFF
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -106,9 +106,11 @@ export interface InputDefinitionBuilder {
   warn(msg: string): void
 }
 
+// prettier-ignore
 export interface OutputDefinitionBlock<TypeName extends string>
-  extends NexusGenCustomOutputMethods<TypeName>,
-    NexusGenCustomOutputProperties<TypeName> {}
+       extends NexusGenCustomOutputMethods<TypeName>,
+               NexusGenCustomOutputProperties<TypeName>
+       {}
 
 /**
  * The output definition block is passed to the "definition"
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -64,28 +64,24 @@ export type NexusOutputFieldDef = NexusOutputFieldConfig<string, any> & {
   subscribe?: GraphQLFieldResolver<any, any>
 }
 
-/**
- * Ensure type-safety by checking
- */
-export type ScalarOutSpread<TypeName extends string, FieldName extends string> = NeedsResolver<
-  TypeName,
-  FieldName
-> extends true
-  ? HasGen3<'argTypes', TypeName, FieldName> extends true
-    ? [ScalarOutConfig<TypeName, FieldName>]
-    : [ScalarOutConfig<TypeName, FieldName>] | [FieldResolver<TypeName, FieldName>]
-  : HasGen3<'argTypes', TypeName, FieldName> extends true
-  ? [ScalarOutConfig<TypeName, FieldName>]
-  : [] | [FieldResolver<TypeName, FieldName>] | [ScalarOutConfig<TypeName, FieldName>]
+// prettier-ignore
+export type ScalarOutSpread<TypeName extends string, FieldName extends string> =
+  NeedsResolver<TypeName, FieldName> extends true
+    ? HasGen3<'argTypes', TypeName, FieldName> extends true
+      ? [ScalarOutConfig<TypeName, FieldName>]
+      : [ScalarOutConfig<TypeName, FieldName>] | [FieldResolver<TypeName, FieldName>]
+    : HasGen3<'argTypes', TypeName, FieldName> extends true
+      ? [ScalarOutConfig<TypeName, FieldName>]
+      : [] | [FieldResolver<TypeName, FieldName>] | [ScalarOutConfig<TypeName, FieldName>]
 
-export type ScalarOutConfig<TypeName extends string, FieldName extends string> = NeedsResolver<
-  TypeName,
-  FieldName
-> extends true
-  ? OutputScalarConfig<TypeName, FieldName> & {
-      resolve: FieldResolver<TypeName, FieldName>
-    }
-  : OutputScalarConfig<TypeName, FieldName>
+// prettier-ignore
+export type ScalarOutConfig<TypeName extends string, FieldName extends string> =
+    NeedsResolver<TypeName, FieldName> extends true
+    ? OutputScalarConfig<TypeName, FieldName> &
+      {
+        resolve: FieldResolver<TypeName, FieldName>
+      }
+    : OutputScalarConfig<TypeName, FieldName>
 
 export type FieldOutConfig<TypeName extends string, FieldName extends string> = NeedsResolver<
   TypeName,
@@ -158,8 +154,7 @@ export class OutputDefinitionBlock<TypeName extends string> {
     // 2. NexusOutputFieldDef is contrained to be be a string
     // 3. so `name` is not compatible
     // 4. and changing FieldOutConfig to FieldOutConfig<string breaks types in other places
-    const field: any = { name, ...fieldConfig }
-    this.typeBuilder.addField(this.decorateField(field))
+    this.typeBuilder.addField(this.decorateField({ name, ...fieldConfig } as any))
   }
 
   protected addScalarField(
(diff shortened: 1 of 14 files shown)


tests/integration/_app.ts
```diff
@@ -96,3 +98,76 @@ export const Mutation = mutationType({
     })
   },
 })
+
+export const Subscription = subscriptionType({
```
jasonkuhrt, line 102: Now we cover subscription type in integration tests.

src/definitions/subscriptionType.ts
```diff
@@ -66,66 +17,92 @@ export interface SubscribeFieldConfig<TypeName extends string, FieldName extends
...
+export interface SubscribeFieldConfig<TypeName extends string, FieldName extends string>
+  extends SubscribeFieldConfigBase<FieldName>,
+          CommonOutputFieldConfig<'Subscription', FieldName>
+  {
+    type: GetGen<'allOutputTypes'> | AllNexusOutputTypeDefs
   }
 
+export interface SubscriptionBuilderInternal extends ObjectDefinitionBuilder<'Subscription'> {}
+
+export class SubscriptionBuilder {
```
jasonkuhrt, line 41: No more subclassing `ObjectDefinitionBuilder`

src/definitions/subscriptionType.ts
```diff
@@ -66,66 +17,92 @@ export interface SubscribeFieldConfig<TypeName extends string, FieldName extends
...
+
   get list() {
     if (this.isList) {
       throw new Error('Cannot chain list.list, in the definition block. Use `list: []` config value')
     }
-    return new SubscriptionDefinitionBlock(this.typeBuilder, true)
+    return new SubscriptionBuilder(this.typeBuilder, true)
+  }
+
+  string<FieldName extends string>(fieldName: FieldName, config: FieldShorthandConfig<FieldName>) {
```
jasonkuhrt, line 51: No more spread confusion needed, as config will _always_ be needed.

src/definitions/subscriptionType.ts
```diff
@@ -66,66 +17,92 @@ export interface SubscribeFieldConfig<TypeName extends string, FieldName extends
...
-    ...opts: ScalarSubSpread<'Subscription', FieldName>
-  ) {
-    this.addScalarField(fieldName, 'String', opts)
+  int<FieldName extends string>(fieldName: FieldName, config: FieldShorthandConfig<FieldName>) {
+    this.fieldShorthand(fieldName, 'Int', config)
   }
 
-  int<FieldName extends string>(fieldName: FieldName, ...opts: ScalarSubSpread<'Subscription', FieldName>) {
-    this.addScalarField(fieldName, 'Int', opts)
+  // prettier-ignore
```
jasonkuhrt, line 59: Much easier to read these signatures as one-liners.

src/definitions/subscriptionType.ts
```diff
@@ -66,66 +17,92 @@ export interface SubscribeFieldConfig<TypeName extends string, FieldName extends
 
   subscribe(
     root: object,
-    args: ArgsValue<TypeName, FieldName>,
+    args: ArgsValue<'Subscription', FieldName>,
```
jasonkuhrt, line 20: Hardcoding instead of needless indirection.

src/definitions/subscriptionType.ts
```diff
@@ -1,63 +1,14 @@
...
-  type: GetGen<'allOutputTypes'> | AllNexusOutputTypeDefs
-
-  /**
-   * Resolve method for the field
-   */
+export interface SubscribeFieldConfigBase<FieldName extends string, Event = any> {
   resolve(
-    root: T,
-    args: ArgsValue<TypeName, FieldName>,
+    root: Event,
```
jasonkuhrt, line 10: We have some feature work to do to make this typesafe in practice for users.

src/definitions/subscriptionField.ts
```diff
@@ -14,7 +14,7 @@ export function subscriptionField<FieldName extends string>(
     type: 'Subscription',
     definition(t) {
       const finalConfig = typeof config === 'function' ? config() : config
-      t.field(fieldName, finalConfig)
+      t.field(fieldName, finalConfig as any)
```
jasonkuhrt, line 17: Gets us around this issue for now:

image

jasonkuhrt: Maybe we should try to fix now.

src/definitions/definitionBlocks.ts
```diff
@@ -158,8 +154,7 @@ export class OutputDefinitionBlock<TypeName extends string> {
     // 2. NexusOutputFieldDef is contrained to be be a string
     // 3. so `name` is not compatible
     // 4. and changing FieldOutConfig to FieldOutConfig<string breaks types in other places
-    const field: any = { name, ...fieldConfig }
```
jasonkuhrt, line 161: Pretty confusing identifier name.

src/definitions/definitionBlocks.ts
```diff
@@ -64,28 +64,24 @@ export type NexusOutputFieldDef = NexusOutputFieldConfig<string, any> & {
...
-
-export type ScalarOutConfig<TypeName extends string, FieldName extends string> = NeedsResolver<
-  TypeName,
-  FieldName
-> extends true
-  ? OutputScalarConfig<TypeName, FieldName> & {
-      resolve: FieldResolver<TypeName, FieldName>
-    }
-  : OutputScalarConfig<TypeName, FieldName>
+// prettier-ignore
```
jasonkuhrt, line 67: Formatting improvements.

jasonkuhrt: This was after I had made actual logical changes, then undid them. Didn't revert everything as the new formatting seemed better.

src/definitions/subscriptionType.ts
Weakky: Isn't it typesafe already?

Weakky:
> It no longer sub-classes from Object Type. This means we lose interface implements support. However, this was never tested before and it is unclear it's even a feature we want.

Technically, it doesn't seem something that 99% of use-cases will need, so I'd be fine rolling this out until someone yells.

However, while this will require slightly more work, I don't see reasons why we could not add GraphQL interface inheritance support even if the internal subscription class no longer "super-classes" the object type one. That's really just an implementation detail.

@tgriesser WDYT?

src/definitions/subscriptionType.ts
jasonkuhrt: No, the resolve `root` type should match the type returned from `subscribe`.

https://github.com/graphql-nexus/schema/issues/454

src/definitions/subscriptionField.ts
jasonkuhrt: Found a fix, solution was to make `extendType` builder API polymorphic over the value of `TypeName`. Specifically branching over if it is `Subscription` or not.

src/definitions/extendType.ts
```diff
@@ -1,11 +1,14 @@
...
 import { AllOutputTypesPossible } from '../typegenTypeHelpers'
 import { OutputDefinitionBlock } from './definitionBlocks'
+import { IsSubscriptionType, SubscriptionBuilder } from './subscriptionType'
 import { NexusTypes, withNexusSymbol } from './_types'
 
 export interface NexusExtendTypeConfig<TypeName extends string> {
   type: TypeName
-  definition(t: OutputDefinitionBlock<TypeName>): void
+  definition(
+    t: IsSubscriptionType<TypeName> extends true ? SubscriptionBuilder : OutputDefinitionBlock<TypeName>
```
jasonkuhrt, line 10: Make `extendType` builder API polymorphic over the value of `TypeName`. Specifically branching over if it is `Subscription` or not. This is needed since the types are different. The naming is a new problem now. That is, `OutputDefinitionBlock` should be called something else. Or, it should be refactored to still fulfill the role of generic handling of `Output`.

src/definitions/subscriptionType.ts
Weakky: Yeah but the `Event` generic already does that, although it seems it's not properly captured as explained by that issue 👍

jasonkuhrt:
I think this is good enough to not keep blocking the fix from being released.

PR #571, opened 2020-10-22
fix: "union too complex to represent" for large union types

### LINKED ISSUE #503: Got error "Expression produces a union type that is too complex to represent" since 0.15.0
Hi,

Since I upgraded `@nexus/schema` from version `0.14.0` to `0.15.0`, I got the following error on one of my union type:

image

I'm using latest version of typescript `4.0.2` and graphql `14.7.0` (also tested with graphql `15.3.0`)

My union type looks like this:
```ts
export const EmbeddedVideoUnion = unionType({
  name: "EmbeddedVideoUnion",
  definition(t) {
    t.members(YoutubeEmbeddedVideo, VimeoEmbeddedVideo);
    t.resolveType(parent => {
      switch (parent.type) {
        case EmbeddedVideoType.YOUTUBE:
          return "YoutubeEmbeddedVideo";
        case EmbeddedVideoType.VIMEO:
          return "VimeoEmbeddedVideo";
      }
      return null;
    });
  },
});
```

The error goes away if I explicitly set the name of the union like this:
```ts
export const EmbeddedVideoUnion = unionType<"EmbeddedVideoUnion">({
 // ...
});
```

But then it triggers similar errors on other unions and interfaces types.

Note that my schema is quite big (~160 types, 160 inputs, 13 unions, 7 interfaces, 80 mutations...) and I'm using ts imports instead of string references in all my nexus types.

I looked at the recent commits and I found this one: 61eccca2004967164ca0dd14037092ca5b7de862 which may be the cause.

Weakky:
Closes #503


src/typegenTypeHelpers.ts
```diff
@@ -157,33 +157,20 @@ export type GetGen<K extends GenTypesShapeKeys, Fallback = any> = NexusGen exten
...
-    ? K extends keyof GenTypes
-      ? K2 extends keyof GenTypes[K]
-        ? GenTypes[K][K2]
-        : any
-      : any
-    : any
-  : any
+  K2 extends Extract<keyof GenTypesShape[K], string>,
+  Fallback = any
+> = K2 extends keyof GetGen<K, never> ? GetGen<K>[K2] : Fallback
```
Weakky, line 162: Don't ask me the _exact_ reason why this is fixing anything. I assume this is due to some TS internals (which may backfire to us later...) and the way they compute union distribution. Perhaps in this case the computation is deferred, perhaps this has no relation at all and I'm saying bs.

A bit of explanation though: We intentionally use `GetGen<K, never>` so that if `GetGen<K>` fails, we fallback to `never` which ensures that `K2 extends of never` will never be true, thus falling in the false case.

tests/integrations/unionTooLargeToRepresent/__app.ts
```diff
@@ -322,14 +322,14 @@ export const types = [
...
-      t.members(...typeNames.slice(50, 150))
+      t.members(...(typeNames as any).slice(50, 150))
       t.resolveType(() => null)
     },
   }),
   unionType({
     name: 'MediumUnionTypes',
     definition(t) {
-      t.members(...typeNames.slice(0, 100))
+      t.members(...(typeNames as any).slice(0, 100))
```
jasonkuhrt, line 332: Is this to make `ts-jest` happy or IDE? In my case VSCode was correctly interpreting the typings here. And given your mentioned solution to make ts-jest ignore these integration tests then I don't think we need these `any` changes here anymore?

Also these `any` would affect the actual integration test too, which isn't ideal.

jasonkuhrt (approved):
Just one comment but otherwise LGTM!!

tests/integrations/unionTooLargeToRepresent/__app.ts
jasonkuhrt: I updated the TS integration tester to use a local tsconfig.json if it is present. This way we can satisfy the IDE and the integration twice at once.

jasonkuhrt:
All tests are passing. But some (one now) fail because of:

image

Ignoring this.

PR #574, opened 2020-10-25
Update code blocks in Chapter 6 of the tutorial to fix error

shreyas44:
For the types in `args` there was `schema.<type>` in all the blocks of code which raises an error. However it should be `stringArg` and `intArg` rather than `schema.stringArg` and `schema.intArg`


jasonkuhrt:
Thanks!

PR #575, opened 2020-10-26
chore(docs): fix nullabillity guard docs

Lilmortal:
onNullGuarded does not exist on `"@nexus/schema": "^0.16.0",`.
Btw the testing is failing because I need some API keys I think.


Lilmortal:
onNullGuarded does not exist on `"@nexus/schema": "^0.16.0",`.
Btw the testing is failing because I need some API keys I think.

jasonkuhrt (approved):
Thanks @Lilmortal!

PR #590, opened 2020-10-30
improve: remove dynamic output builtins

### LINKED ISSUE #581: Drop Dynamic methods
They have since been replaced by the connections plugin. This will be technically a breaking change but few users should be affected.

jasonkuhrt:
close #581


Weakky (approved):
Apart from tests not passing (haven't review that), changes look fine to me!

PR #592, opened 2020-10-30
improve: remove resolver shorthands

### LINKED ISSUE #582: Drop the resolver shorthand
We all agree that it is rarely useful:

- Fields requiring anything even `description` will need object style
- Subscription fields can't support it because they need `subscribe` config field (Note: @Weakky pointed out that the shorthand form could be appropriated in this case to be the subscribe method since in this case the `resolve` method can be inferred as a passthrough. This of course has its own tradeoffs too, e.g. looks-the-same-but-isn't-the-same API "trickery" (arguably))
-

jasonkuhrt:
closes #582

As discussed. In most cases resolver shorthands are not a serious use of the API.

BREAKING CHANGE:

Resolver shorthand API is now removed. The following will now not typecheck:

```ts
t.string('foo', () => ... )
```

Instead use:

```ts
t.string('foo', { resolve: () => ... })
```

Runtime support is still intact but will result in a logged warning. Runtime support will be removed in the next Nexus release.

--- 

I probably missed areas of the codebase that can be simplified as a result. In particular the parts in the builder module that need to do runtime checks. Also maybe something about typegen?

### DIFF
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -162,7 +162,7 @@ export class OutputDefinitionBlock<TypeName extends string> {
       this.decorateField({
         name: fieldName,
         type: typeName,
-        ...opts[0],
+        ...(opts[0] as any),
       })
     )
   }
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -154,19 +154,21 @@ export class OutputDefinitionBlock<TypeName extends string> {
     typeName: BaseScalars,
     opts: [] | ScalarOutSpread<TypeName, any>
   ) {
-    if (typeof opts[0] === 'function') {
-      console.warn(
-        `Since v0.18.0 Nexus no longer supports resolver shorthands like:\n\n    t.string("${fieldName}", () => ...).\n\nInstead please write:\n\n    t.string("${fieldName}", { resolve: () => ... })`
-      )
+    let config: NexusOutputFieldDef = {
+      name: fieldName,
+      type: typeName,
     }
 
-    this.typeBuilder.addField(
-      this.decorateField({
-        name: fieldName,
-        type: typeName,
-        ...(opts[0] as any),
-      })
-    )
+    if (typeof opts[0] === 'function') {
+      config.resolve = opts[0] as any
+      console.warn(
+        `Since v0.18.0 Nexus no longer supports resolver shorthands like:\n\n    t.string("${fieldName}", () => ...).\n\nInstead please write:\n\n    t.string("${fieldName}", { resolve: () => ... })\n\nIn the next version of Nexus this will be a runtime error.`
+      )
+    } else {
+      config = { ...config, ...opts[0] }
+    }
+
+    this.typeBuilder.addField(this.decorateField(config))
   }
 
   protected decorateField(config: NexusOutputFieldDef): NexusOutputFieldDef {
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -58,13 +58,15 @@ export type NexusOutputFieldDef = NexusOutputFieldConfig<string, any> & {
 
 // prettier-ignore
 export type ScalarOutSpread<TypeName extends string, FieldName extends string> =
-  HasGen3<'argTypes', TypeName, FieldName> extends true
+  NeedsResolver<TypeName, FieldName> extends true
     ? [ScalarOutConfig<TypeName, FieldName>]
-    : [ScalarOutConfig<TypeName, FieldName>] | []
+    : HasGen3<'argTypes', TypeName, FieldName> extends true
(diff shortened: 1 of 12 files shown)


tests/typegen/types.gen.ts
```diff
@@ -10,206 +10,220 @@ declare global {
...
   CreatePostInput: {
     // input type
-    author: string // ID!
-    geo: Array<Array<number | null>> // [[Float]!]!
-    name: string // String!
-  }
+    author: string; // ID!
+    geo: Array<Array<number | null>>; // [[Float]!]!
+    name: string; // String!
+  };
```
Weakky, line 16: Would be great if you could prettify that file so that there's no unnecessary diffs

jasonkuhrt:
Just noticed the tests are failing 😅 will fix

src/definitions/definitionBlocks.ts
```diff
@@ -58,13 +58,9 @@ export type NexusOutputFieldDef = NexusOutputFieldConfig<string, any> & {
 
 // prettier-ignore
 export type ScalarOutSpread<TypeName extends string, FieldName extends string> =
-  NeedsResolver<TypeName, FieldName> extends true
-    ? HasGen3<'argTypes', TypeName, FieldName> extends true
-      ? [ScalarOutConfig<TypeName, FieldName>]
-      : [ScalarOutConfig<TypeName, FieldName>] | [FieldResolver<TypeName, FieldName>]
-    : HasGen3<'argTypes', TypeName, FieldName> extends true
-      ? [ScalarOutConfig<TypeName, FieldName>]
-      : [] | [FieldResolver<TypeName, FieldName>] | [ScalarOutConfig<TypeName, FieldName>]
```
Weakky, line 67: Could you explain the reason why we no longer need the `NeedsResolver` type?

jasonkuhrt: Ah, I think we should be keeping that you're right. When `NeedsResolver` is true then `[]` should never be an option.

It seems we don't have any test coverage for this case... I'll see if I can add some.

tests/typegen/types.gen.ts
tgriesser: Was going to say the same, thought that should already happen pre-commit.

tgriesser (changes requested):
Can we add temporary a runtime check for this if we see a fn as the second arg to a field to log a console error to warn folks about this change - in-case they're not being super type-safe when using nexus. 

I could see it being easy to miss if things started up as normal - not sure what the behavior would be actually.

tests/typegen/types.gen.ts
tgriesser: Oh, need to change https://github.com/graphql-nexus/schema/blob/2de6f8971e5caac24af416342708c05218edc142/tests/typegenPrinter.spec.ts#L17-L18 to ```prettierConfig: path.join(__dirname, '../.prettierrc'),```

jasonkuhrt:
@tgriesser could you point me to the builder code. I actually couldn't find it the first time around when trying to remove runtime support for it.

Weakky:
@jasonkuhrt here https://github.com/graphql-nexus/schema/blob/9bfdf2cdc55d962ba89422f14ec4ecb11ec2bfa2/src/definitions/definitionBlocks.ts#L154-L170

src/definitions/definitionBlocks.ts
```diff
@@ -156,17 +152,19 @@ export class OutputDefinitionBlock<TypeName extends string> {
...
-    let config: NexusOutputFieldDef = {
-      name: fieldName,
-      type: typeName,
-    }
     if (typeof opts[0] === 'function') {
-      // FIXME ditto to the one in `field` method
-      config.resolve = opts[0] as any
-    } else {
-      config = { ...config, ...opts[0] }
+      console.warn(
```
tgriesser, line 156: Should this be a `console.error` rather than a warn? If we're going to gracefully degrade (which I don't think is a bad option, to remove the types but support the existing runtime behavior) then I think it'd make sense as a `warn`. Otherwise I think it should be a `console.error`.

jasonkuhrt: > (which I don't think is a bad option, to remove the types but support the existing runtime behavior)

Ah that's a great point, sounds good to me. I'll mention in the log that this will be a hard runtime error in the next release.

jasonkuhrt:
@Weakky or @tgriesser any idea what's up with the failing tests? Not from this PR looks like. Trying to fix anyways.

jasonkuhrt:
Hey @Weakky and @tgriesser ready for re-review I think!

Weakky:
@jasonkuhrt I think the CI errors are related to @tgriesser's last commit, but I can't fix it either. Even after updating the snapshots, the errors persist :(

Weakky (approved):
Looks good to me! 👍

jasonkuhrt:
@Weakky ok we'll deal with it later, the tests were passing locally before so let's assume that's enough here.

PR #598, opened 2020-11-02
fix: #588, #384 non-null list items & connection nullability

### LINKED ISSUE #384: Ability to specify nullability of connectionPlugin output
Right now there's no way to specify whether or not a connection field is nullable via the plugin. It seems to follow the value of the schema's `nonNullDefaults.output` setting right now.

It'd be nice to able to specify the nullability of both the connection and the inner `edges` (and `nodes`) fields. This seems to matter more when `nonNullDefaults.output` is `false`, since it leads to types like:

```graphql
type SomeType {
  someConnection: SomeConnection
}

type SomeConnection {
  edges: [SomeEdge]
  nodes: [SomeNode!]
  pageInfo: PageInfo!
}

type SomeEdge {
  node: SomeNode
  cursor: String!
}
```

This forces the client to check for nullability on `someConnection`, `someConnection.edges`, and `someConnection.edges[0]`.

If anyone has guidelines on what a good solution would be here, I'd be happy to send a PR 😄

### LINKED ISSUE #588: `t.list.field` doesn't obey `nonNullDefaults`
The following definition will result in the subsequent schema with `nonNullDefaults` set to `{ output: true, input: true }`:
```typescript
queryType({
  definition(t) {
    t.list.int('ints')
  }
})
```
```graphql
type Query {
  ints: [Int]!
}
```

Instead, [according to the documentation](https://nexusjs.org/docs/guides/schema#lists-and-nullability), it should result in this schema:
```graphql
type Query {
  ints: [Int!]!
}
```

The alternative is not as ergonomic:
```typescript
queryType({
  definition(t) {
    t.int('ints', { list: [true] })
  }
})
```

tgriesser:
Fixes #588 by ensuring that list items follow `nonNullDefaults`, which was lost in the #508 refactor

Fixes #384 by removing explicit nullability config from the connection `edges` definition, and also allowing `nonNullDefaults` to be supplied for the connection "types" generated by the creation of a connection field. Allows you to configure both globally in the connection plugin field config, and when the connection field is defined.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -1217,15 +1217,22 @@ export class SchemaBuilder {
     isNonNull: boolean
   ): T {
     if (list) {
-      type = this.decorateList(type, list)
+      type = this.decorateList(type, list, isNonNull)
     }
     return (isNonNull ? GraphQLNonNull(type) : type) as T
   }
 
-  protected decorateList<T extends GraphQLOutputType | GraphQLInputType>(type: T, list: true | boolean[]): T {
+  protected decorateList<T extends GraphQLOutputType | GraphQLInputType>(
+    type: T,
+    list: true | boolean[],
+    isNonNull: boolean
+  ): T {
     let finalType = type
     if (!Array.isArray(list)) {
-      return GraphQLList(type) as T
+      if (isNonNull) {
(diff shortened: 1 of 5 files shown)


jasonkuhrt (approved):
LGTM

PR #599, opened 2020-11-02
feat: interfaces implementing interfaces can resolveType w/ interface name

tgriesser:
Will probably also try to upstream this PR's behavior to `graphql-js`, but it seems to make sense that if you have an `interface` which implements another `interface`, in the `resolveType` of the parent you should be able to `resolveType` an implementing interface's `__typename`, and it'd then cascade into that interface type's `resolveType` to determine the concrete type.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -39,6 +39,7 @@ import {
   isUnionType,
   isWrappingType,
   printSchema,
+  defaultTypeResolver,
 } from 'graphql'
 import { arg, ArgsRecord, NexusArgConfig, NexusArgDef } from './definitions/args'
 import {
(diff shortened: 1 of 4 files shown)


tgriesser:
I suppose this would also require a change in the types, to allow the interface be returned from the `resolveType`. Also ping'ed Ivan to see if this is a reasonable idea that might be upstream'ed before implementing here.

jasonkuhrt:
Could you share a minimal code example of the code-reuse your envisioning?

tgriesser:
Nevermind, after discussing w/ Ivan - interfaces implementing inteerfaces

> was added purely to support changing return values of polymorphic fields

and this sort of proper interface chain resolution is out of scope. @jasonkuhrt I was looking to make it so `resolveType` of an `interface` type could return the name of another `interface` which implements the first `interface`, and have the `resolveType` cascade until it found a concrete type. Going to skip this for now though, since this isn't GraphQL's behavior

jasonkuhrt:
@tgriesser Ah I see, that's interesting IMO. Might be worth a new issue for the community to discuss!

PR #602, opened 2020-11-03
feat: improve abstract types

### LINKED ISSUE #188: Union's resolveType doesn't know about __typename

Weakky:
Closes #188

I'm not super familiar with how interfaces implementing other interface work right now, so I'm not sure whether this PR misses something about that feature.

Right now, this PR:

- (BREAKING): move `resolveType` in the type config (so we can have static type errors)
- introduces `isTypeOf` in the type config
- introduces a global config to choose which strategy you want to use to discriminate types (resolveType, isTypeOf, or using the `__typename` discriminator)
- provides type-safety if you want to discriminate your types using __typename
- (BREAKING): default discriminating strategy is: `isTypeOf`. We can revert it to be `resolveType` by default if needed.

**Limitations**:

Ideally, when `{ abstractTypes: { __typename: true, isTypeOf: true, resolveType: true } }` and the user chooses in some places to use the `__typename` strategy, it's only as soon as they've implemented the resolvers with some `__typename` discriminator as return type that the type errors for missing `isTypeOf` and `resolveType` should be gone.

Unfortunately, because we cannot infer whether a resolver returns some `__typename` discriminator from the location of an abstract type, as soon as `abstractTypes: { __typename: true }`, `resolveType` and `isTypeOf` have to become optional.

The only way for us to make `resolveType` and `isTypeOf` optional _only_ when all the field resolvers of types that are part of abstract types are returning a `__typename` property, would be to make fields of a type inferrable and references to a type be only via the values and not via strings, similar to how https://github.com/sikanhe/gqtx works. Otherwise, I don't see any way around this tradeoff.

TODO after merging:

- create shortlinks 
  - [ ] https://nxs.li/guides/abstract-types
  - [ ] https://nxs.li/guides/abstract-types/modular-strategy
  - [ ] https://nxs.li/guides/abstract-types/centralized-strategy
  - [ ] https://nxs.li/guides/abstract-types/discriminant-model-field-strategy
  - [ ] https://nxs.li/guides/abstract-types/using-multiple-strategies
- [x] implement `abstractTypeRuntimeChecks` setting
- [x] tests


src/typegenPrinter.ts
```diff
@@ -140,6 +143,9 @@ export class TypegenPrinter {
         `  allNamedTypes: NexusGenTypes['allInputTypes'] | NexusGenTypes['allOutputTypes']`,
         `  abstractTypes: NexusGenTypes['interfaceNames'] | NexusGenTypes['unionNames'];`,
         `  abstractResolveReturn: NexusGenAbstractResolveReturnTypes;`,
+        `  isTypeOfObjectNames: NexusGenIsTypeOfObjectNames;`,
+        `  resolveTypeImplemented: NexusGenResolveTypeImplemented;`,
+        `  checks: NexusGenChecksConfig;`,
```
Weakky, line 148: Would love better namings...

src/typegenPrinter.ts
```diff
@@ -141,6 +144,9 @@ export class TypegenPrinter {
         `  allNamedTypes: NexusGenTypes['allInputTypes'] | NexusGenTypes['allOutputTypes']`,
         `  abstractTypes: NexusGenTypes['interfaceNames'] | NexusGenTypes['unionNames'];`,
         `  abstractResolveReturn: NexusGenAbstractResolveReturnTypes;`,
+        `  isTypeOfObjectNames: NexusGenIsTypeOfObjectNames;`,
+        `  resolveTypeImplemented: NexusGenResolveTypeImplemented;`,
+        `  features: NexusGenFeaturesConfig;`,
```
Weakky, line 149: Happy to get suggestions for better namings...

src/utils.ts
```diff
@@ -324,6 +331,62 @@ export function assertNoMissingTypes(schema: GraphQLSchema, missingTypes: Record
...
+          `${kind} "${
+            type.name
+          }" is either missing a \`resolveType\` implementation or its members are missing an \`isTypeOf\` implementation: ${typesWithoutIsTypeOf
+            .map((t) => `"${t.name}"`)
+            .join(', ')}`
+        )
+      )
+    }
+  })
+}
```
Weakky, line 388: Not sure whether these guards shouldn't be throws instead of console.error's. The previous guard on resolveType was a console.error, but I suppose that was mainly to not prevent typegen from running.
As this guard is run _after_ typegen is run, it shouldn't block anything critical.

Is there any reason why we wouldn't wanna throw?

src/typegenPrinter.ts
jasonkuhrt: - `objectsUsingAbstractStrategyIsTypeOf`
- `objectsUsingAbstractStrategyResolveType`

src/typegenPrinter.ts
```diff
@@ -345,6 +351,46 @@ export class TypegenPrinter {
     return `export type ${exportName} = ${typeDef};`
   }
 
+  printIsTypeOfObjectNames(exportName: string) {
+    const obj = this.groupedTypes.object.filter((o) => o.isTypeOf !== undefined)
```
jasonkuhrt, line 355: `objectTypes`?

src/typegenPrinter.ts
```diff
@@ -345,6 +351,46 @@ export class TypegenPrinter {
...
+        ? 'never'
+        : obj
+            .map((o) => JSON.stringify(o.name))
+            .sort()
+            .join(' | ')
+    return `export type ${exportName} = ${typeDef};`
+  }
+
+  printResolveTypeUnionInterface(exportName: string) {
+    const obj = [...this.groupedTypes.interface, ...this.groupedTypes.union].filter(
```
jasonkuhrt, line 367: `abstractTypes`?

src/definitions/_types.ts
```diff
@@ -163,3 +163,11 @@ export type NexusGraphQLInterfaceTypeConfig = WithExt<
 export type NexusGraphQLSchema = Omit<GraphQLSchema, 'extensions'> & {
   extensions: { nexus: NexusSchemaExtension }
 }
+
+export interface NexusFeatures {
```
jasonkuhrt, line 167:
```suggestion
/**
 * Toggle various features of Nexus on or off
 */
export interface NexusFeatures {
```

src/definitions/_types.ts
```diff
@@ -163,3 +163,11 @@ export type NexusGraphQLInterfaceTypeConfig = WithExt<
...
+  abstractTypes?: {
```
jasonkuhrt, line 168:
```suggestion
  /**
   * Toggle abstract type implementation strategies. For more info please refer to to the [abstract types guide](https://nxs.li/guides/abstract-types)
   */
  abstractTypes?: {
```

We can do a lot more jsdoc work here and on each strategy as well. Bit hard to do that via code suggest though.

What I would say is that we can copy the content from each strategy section in the abstract types guide into jsdoc here.

Kind of intense but going to be pretty helpful I bet.

src/builder.ts
```diff
@@ -229,6 +231,7 @@ export interface BuilderConfig {
    * Otherwise, uses `printSchema` from graphql-js
    */
   customPrintSchemaFn?: typeof printSchema
+  features?: NexusFeatures
```
jasonkuhrt, line 234: If we put jsodc on `NexusFeatures` will that jsdoc apply to the `features` property here too if it has none?

src/builder.ts
```diff
@@ -1581,14 +1579,35 @@ export function makeSchemaInternal(config: SchemaConfig) {
   return { schema, missingTypes, finalConfig }
 }
 
+function setConfigDefaults(config: SchemaConfig): SchemaConfig {
```
jasonkuhrt, line 1582: Should we have two types here? E.g.: `SchemaConfigInput` and `SchemaConfig`

src/builder.ts
```diff
@@ -1581,14 +1579,35 @@ export function makeSchemaInternal(config: SchemaConfig) {
...
+  const unionsDefault: NexusFeatures['abstractTypes'] = {
```
jasonkuhrt, line 1583: `abstractTypesDefault`

src/definitions/interfaceType.ts
```diff
@@ -31,23 +31,16 @@ export type NexusInterfaceTypeConfig<TypeName extends string> = {
    * Root type information for this type
    */
   rootTyping?: RootTypingDef
-}
+} & ResolveType<TypeName>
```
jasonkuhrt, line 34: Would `ResolveType` make more sense named `TypeConfigResolveType`?

src/definitions/objectType.ts
```diff
@@ -63,7 +63,8 @@ export type NexusObjectTypeConfig<TypeName extends string> = {
    * Root type information for this type
    */
   rootTyping?: RootTypingDef
-} & NexusGenPluginTypeConfig<TypeName>
+} & IsTypeOf<TypeName> &
```
jasonkuhrt, line 66: Like other comment, here named `TypeConfigIsTypeOf`?

jasonkuhrt (commented):
This is a half review as I had to step out. Will finish more tonight. So far just got to superficial things. Looking forward to more meaty parts later!

src/typegenPrinter.ts
```diff
@@ -345,6 +351,46 @@ export class TypegenPrinter {
...
+      obj.length === 0
+        ? 'never'
+        : obj
+            .map((o) => JSON.stringify(o.name))
+            .sort()
+            .join(' | ')
+    return `export type ${exportName} = ${typeDef};`
+  }
+
+  printResolveTypeUnionInterface(exportName: string) {
```
jasonkuhrt, line 366: `printResolveTypeAbstractTypes`?

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+      resolveType?: AbstractTypeResolver<TypeName>
+    }
+  : {
+      /**
+       * Optionally provide a custom type resolver function. If one is not provided,
+       * the default implementation will call `isTypeOf` on each implementing
+       * Object type.
+       */
+      resolveType: AbstractTypeResolver<TypeName>
+    }
```
jasonkuhrt, line 407: Nice set of building block funcs. All of this could go under an `AbstractTypes` namespace (see later comment).

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+ * Discriminate a type with a __typename: TypeName field
+ */
+type DiscriminateImpl<TypeName extends string, Type> = Type extends Promise<infer A>
+  ? Promise<DiscriminateImpl<TypeName, A>>
+  : Type extends Array<infer B>
+  ? Array<DiscriminateImpl<TypeName, B>>
+  : Type extends ReadonlyArray<infer C>
+  ? ReadonlyArray<DiscriminateImpl<TypeName, C>>
+  : Type extends object
+  ? Type & { __typename: TypeName }
```
jasonkuhrt, line 457: IIRC there is somewhere else in the codebase we do similar or probably same unwrapping of types to get at the actual returned model type. I believe this is pretty fragile logic and should be centralized. If we are duplicating this logic think it might be interesting to have a helper like: 

```ts
type DiscriminateImpl<ModelType, TypeName extends string> = IntersectModelTypeWith<ModelType, { __typename: TypeName }>

type IntersectModelTypeWith<ModelType, T> = ...all the conditional recursive logic... finally... ModelType & T
```

This also helps separate the "business logic" from the lower-level details about how to drill down to the model type. Thus in the end this is a lot easier to read/declarative: `IntersectModelTypeWith<ModelType, { __typename: TypeName }>` IMO

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+      : false
+    : ReturnTypeName extends GetGen<'abstractTypes'> // else if return type is an abstract type
+    ? ShouldDiscriminateAbstractTypeField<ReturnTypeName> extends true // call sub function
+      ? true
+      : false
+    : false
+  : false
+
+/**
+ * Discriminate a type with a __typename: TypeName field
```
jasonkuhrt, line 448:
```suggestion
 * Apply the discriminant strategy to a model type. This adds the __typename: "<TypeNameHere>" property to the model type.
```

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+  ? Type & { __typename: TypeName }
+  : Type
+
+/**
+ * Discriminate a type with a __typename: TypeName field
+ */
+export type Discriminate<TypeName extends string, Type> = TypeName extends
+  | GetGen<'objectNames'>
+  | GetGen<'interfaceNames'>
+  | GetGen<'enumNames'> // if typename is output type
```
jasonkuhrt, line 466:
```
IsOutputType<TypeName> extends true
```

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+  ? DiscriminateImpl<TypeName, Type> // inject __typename discriminator
+  : TypeName extends GetGen<'abstractTypes'> // if typename is abstract type (meaning that `Type` can be a union)
```
jasonkuhrt, line 468: `AbstractTypes.IsAbstractType<TypeName> extends true`

jasonkuhrt (commented):
A night behind here. And I still couldn't get through it all. I hope my feedback isn't too detail oriented. I find that this system is tricky and getting the code readability as high as we can will make a big difference for maintainability. I know many of my points are potentially highly subjective but hopefully they are useful to all.

More feedback tomorrow night if this isn't merged by then :P

jasonkuhrt:
> Unfortunately, because we cannot infer whether a resolver returns some __typename discriminator from the location of an abstract type, as soon as abstractTypes: { __typename: true }, resolveType and isTypeOf have to become optional.

> The only way

For completeness sake I would add that another way is to build a little program that analyzes the user's project and augments typegen with the info we need.

This technically could be done within Nexus transparently to the user.

It would however:

1. Make `typescript` a prod dep of Nexus
2. Slow down typegen significantly
3. Be non-trivial to implement (but after experience with framework work on `addToContext` work I think doable)
4. Make the codebase more complex and harder to maintain
5. Not work in 100% of cases (e.g. abstractions in app code or third party package code that break AST traversal)

src/utils.ts
jasonkuhrt: Yeah throwing but seeing as this isn't an operational error (its user-land) there's little value IMO in having a stack trace reported. I guess since we're only a library the best we can do is signal this category of error for upstream to better filter on.

So for example `throw new ApplicationError(...)` which would include some structured data like:

```
code: 'NEXUS_ERR_ABSTRACT_TYPES_MISSING_RESOLVE_TYPE'
```

Known error cases shouldn't generally have stack traces presented to users is my thinking.

tests/__helpers/testApp.ts
```diff
@@ -43,7 +43,12 @@ export const testApp = (settings: Settings) => {
         const prettierConfigPath = require.resolve('../../.prettierrc')
         const content = await typegenFormatPrettier(prettierConfigPath)(source, type)
 
-        return content.replace("'@nexus/schema'", "'../../..'")
+        return content.replace("'@nexus/schema'", "'../../../src'")
+      },
+      features: {
+        abstractTypes: {
+          resolveType: true,
```
jasonkuhrt, line 50: Any particular reason we don't want to use the default for the integration tests?

jasonkuhrt: This just made me aware of some ambiguity in the settings. I don't think it is documented. Which is: Is this a merge semantic or an overwrite semantic where all the other defaults are `false`?

That is:

```ts
// default
{
  abstractTypes: {
    isTypeOf: true,
    __typename: false,
    resolveType: false,
  }
}
```
Now given input:
```ts
{
  abstractTypes: {
    resolveType: true,
  }
}
```
Does that mean (A):
```ts
{
  abstractTypes: {
    isTypeOf: true,
    __typename: false,
    resolveType: true,
  }
}
```
Or (B):
```ts
{
  abstractTypes: {
    isTypeOf: false,
    __typename: false,
    resolveType: true,
  }
}
```

Overall I like that with B its "what you see is what you get". I can see though how this wouldn't generalize to all cases of settings. And thus would A be surprising behaviour? (Sure we'd mention it in jsdoc...)

tests/plugin.spec.ts
```diff
@@ -211,14 +221,16 @@ describe('plugin', () => {
       types: [
         interfaceType({
           name: 'Node',
+          resolveType(n) {
```
jasonkuhrt, line 224: Redundant?

jasonkuhrt: Also how is this allowed if the feature has not been enabled?

jasonkuhrt (commented):
Finally done!

Great work @Weakky 💯

docs/content/014-guides/021-abstract-types.md
```diff
@@ -0,0 +1,499 @@
...
+            __typename,
+          }
+        })
+      },
+    })
+  },
+})
+```
+
+As you can see the technique looks quite similar at face value to the previous `resolveType` one. However your implementation might not look like this at all. For example maybe your data models already contain you a discriminat property `typeName`. For example:
```
Sytten, line 215: Typo: `you a discriminat`

src/typegenPrinter.ts
Weakky: `objectsUsingAbstractStrategyResolveType` -> `abstractsUsingStrategyResolveType`

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+ */
+export type ShouldDiscriminateResultValue<
+  TypeName extends string,
+  FieldName extends string,
+  ReturnTypeName extends string = GetGen3<'fieldTypeNames', TypeName, FieldName>
+> = GetGen3<'features', 'abstractTypes', '__typename', false> extends true // if feature is enabled
+  ? ReturnTypeName extends GetGen<'objectNames'> | GetGen<'interfaceNames'> | GetGen<'enumNames'> // and return type is either an object type || interface || enum
+    ? ShouldDiscriminateOutputTypeField<ReturnTypeName> extends true // call sub function
+      ? true
+      : false
```
Weakky, line 439: remove this, only fields that return an abstract type should "be discriminated"

src/typegenPrinter.ts
jasonkuhrt: Did this

src/builder.ts
jasonkuhrt: Unfortunately it does not seem to.

src/utils.ts
```diff
@@ -324,6 +331,62 @@ export function assertNoMissingTypes(schema: GraphQLSchema, missingTypes: Record
...
+        new Error(
+          `${kind} "${
+            type.name
+          }" is either missing a \`resolveType\` implementation or its members are missing an \`isTypeOf\` implementation: ${typesWithoutIsTypeOf
+            .map((t) => `"${t.name}"`)
+            .join(', ')}`
+        )
+      )
+    }
+  })
```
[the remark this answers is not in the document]
jasonkuhrt: New issue https://github.com/graphql-nexus/schema/issues/616

jasonkuhrt:
Some discussion on the settings schema led to us concluding in favour of `10b`
```ts
type settings10b = {
  features: {
    abstractTypeRuntimeChecks: boolean
    abstractTypeStrategies: {
      __typename: boolean
      isTypeOf: boolean
      resolveType: boolean
    }
  }
}
```

All:

```ts
type settings0 = {
  features: {
    abstractTypes: {
      validateImplementations: boolean
      discriminantModelPropertyStrategy: boolean
      modularStrategy: boolean
      centralizedStrategy: boolean
    }
  }
}

type settings1 = {
  features: {
    abstractTypeRuntimeChecks: boolean
    abstractTypeStrategies: {
      discriminantModelProperty: boolean
      modular: boolean
      centralized: boolean
    }
  }
}

type settings2 = {
  features: {
    abstractTypes: {
      runtimeCheckers: boolean
      strategies: {
        // discriminantModelProperty: boolean
        // modular: boolean
        // centralized: boolean
        //
        discriminantModelProperty: boolean
        predicateOnObjectType: boolean
        discriminatorOnAbstractType: boolean
      }
    }
  }
}

type settings3 = {
  features: {
    abstractTypes: {
      runtimeCheckers: boolean
      strategies: {
        __typename: boolean
        isTypeOf: boolean
        resolveType: boolean
      }
    }
  }
}

type settings4 = {
  features: {
    abstractTypesRuntimeChecks: boolean
    abstractTypesStrategyDiscriminantModelProperty: boolean
    abstractTypesStrategymodularStrategy: boolean
    abstractTypesStrategycentralizedStrategy: boolean
  }
}

// ----------------------------------------------------------------------

type settings10a = {
  features: {
    abstractTypeRuntimeChecks: boolean
    abstractTypeStrategies: {
      discriminantModelProperty: boolean
      modular: boolean
      centralized: boolean
    }
  }
}

type settings10b = {
  features: {
    abstractTypeRuntimeChecks: boolean
    abstractTypeStrategies: {
      __typename: boolean
      isTypeOf: boolean
      resolveType: boolean
    }
  }
}

```

src/utils.ts
jasonkuhrt: It now throws in production otherwise logs

src/typegenTypeHelpers.ts
```diff
@@ -255,3 +260,221 @@ export type NeedsResolver<TypeName extends string, FieldName extends string> = H
...
+  : true // otherwise enable __typename feature
+
+/**
+ * Returns whether a field which type is a union should discriminate its return type with __typename
+ */
+export type ShouldDiscriminateResultValue<
+  TypeName extends string,
+  FieldName extends string,
+  ReturnTypeName extends string = GetGen3<'fieldTypeNames', TypeName, FieldName>
+> = GetGen3<'features', 'abstractTypes', '__typename', false> extends true // if feature is enabled
```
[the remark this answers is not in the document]
jasonkuhrt: Done, `IsFeatureEnabled2` until we're using TS 4.0

src/utils.ts
```diff
@@ -331,10 +331,8 @@ export function assertNoMissingTypes(schema: GraphQLSchema, missingTypes: Record
...
 }
 
-export function assertAbstractTypesCanBeDiscriminated(schema: NexusGraphQLSchema, features: NexusFeatures) {
-  // If backing type check is enabled, we can no longer know for sure if a type is properly discriminated
-  // since it could be discriminated via the `__typename` field in resolvers, which we can't check at runtime
-  if (features.abstractTypeStrategies.__typename === true) {
+export function runAbstractTypeRuntimeChecks(schema: NexusGraphQLSchema, features: NexusFeatures) {
+  if (!features.abstractTypeRuntimeChecks) {
     return
   }
```
Weakky: I think the `__typename` strategy should still be disabling the runtime checks, even if that's confusing. Otherwise, we'll have false reports

examples/apollo-fullstack/src/fullstack-typegen.ts
```diff
@@ -196,7 +196,7 @@ export interface NexusGenArgTypes {
   }
 }
 
-export interface NexusGenAbstractResolveReturnTypes {}
+export interface NexusGenAbstractTypesMapResolveTypeMethodReturnType {}
```
Weakky, line 199: Not sure I like this naming any better, it's super hard to read

src/utils.ts
jasonkuhrt: Yeah thought about it. Should the source of truth be the config or arbitrary logic anywhere in the coebase (don't mean that negatively)?

examples/apollo-fullstack/src/fullstack-typegen.ts
jasonkuhrt: Do you feel one of these makes it easier or harder to understand the intent?

jasonkuhrt (approved):
LGTM! (just need to work through the todo list)

examples/apollo-fullstack/src/fullstack-typegen.ts
Weakky: I think it makes it harder to understand the intent. How about something like `NexusGenAbstractTypeMembers` ?

src/utils.ts
Weakky: Ohhhh, I did not see this was handled in the `setConfigDefaults` function. Yeah, I frankly don't know the best approach here

Weakky: I'll leave it the way you did!

jasonkuhrt: Dumb settings layer and smart components or smart settings layer and dumb components.

I think I'm in favour of smart settings layer because it can be thought of as an integration layer. Each component then doesn't need to know about each other. If there are rules about how they integrate, that can be done higher level rather than violating the logical encapsulation rules.

This is actually one of the ideas driving `setset`.

Once settings have been resolved, the final config is a kind of "absolute source of truth" that drives the rest of the system.

This design is arguably more functional in the sense that it is closer to an assembly line of transformations.

examples/apollo-fullstack/src/fullstack-typegen.ts
jasonkuhrt: I think it'll be hard to find the perfect thing so fine with me. 

I'm not actually sure we wanted to encode that these types are the method return type or not come to think of it... 🤔 

If not your name idea makes sense especially.

I'm not sure we care about encoding the fact that this is a lookup/map/whatever data structure to go from an abstract type to its members. It might be useful to have a naming pattern for all such data structures. E.g.: `<A>To<B>` or `<A>Map<B>`. Looking at the dozen or two Nexus gen fields it feels some naming conventions could help.

jasonkuhrt:
Spun out of this PR: https://github.com/graphql-nexus/schema/issues/623

Weakky:
Note: We decided to inject the `__typename` discriminant directly in the typegen file to avoid messy types because of TS conditional transformations. Here's an example:

image

The produced unions are now much much cleaner.

jasonkuhrt:
Great job @Weakky !!!

jasonkuhrt:
Shortlinks have been updated!

PR #604, opened 2020-11-03
feat: Allow specifying a node module for type imports

tgriesser:
Previously, when specifying `rootTyping.path`, on a scalar it needed to be an absolute path. 

This allows you to instead specify a named `node_module`, which is often more useful in monorepo type projects.


jasonkuhrt (approved):
Nice one, LGTM

PR #606, opened 2020-11-04
feat: Allow contextType to specify an TypingImport config

### LINKED ISSUE #265: Find more seamless way (ideally zero-config) to register Context type
The following is the current way to get a typed context object using Nexus. This feels pretty cumbersome and I'd prefer a zero-config way (while still being transparent) to "register" the Context type.

image

tgriesser:
Closes #265

Allows specifying the `contextType` explicitly, similar to the `rootTyping` for an individual type.


src/builder.ts
```diff
@@ -262,9 +263,13 @@ export interface TypegenInfo {
...
   backingTypeMap: { [K in GetGen<'objectNames'>]?: string }
   /**
-   * The type of the context for the resolvers
+   * The string type of the context for the resolvers
    */
   contextType?: string
+  /**
+   * Info about where to import the context from
+   */
+  contextTypeImport?: TypingImport
```
Weakky, line 272: Any reason to have two different fields? I feel like this is even more confusing as it's not clear which one you're supposed to pick.

Would you consider making `contextType` independent of `typegenAutoConfig.sources`, therefore only allowing `contextType: TypingImport`? From my experience, the context type is 99% of the case in a separate file/folder than where your models lives anyway.

tgriesser: > Any reason to have two different fields? I feel like this is even more confusing as it's not clear which one you're supposed to pick.

I thought this was only an internal type (`TypegenInfo` vs `TypegenAutoConfigOptions` which doesn't have this), so it didn't matter.

> Would you maybe consider making contextType entirely independent of typegenAutoConfig.sources, therefore only allowing contextType: TypingImport? From my experience, the context type is 99% of the case in a separate file/folder than where your models live anyway. It always felt weird to configure the context type as part of your backing/root types, which are arguably two different things.

Yeah that sounds reasonable as well!

jasonkuhrt: > Would you maybe consider making contextType entirely independent of typegenAutoConfig.sources, therefore only allowing contextType: TypingImport? From my experience, the context type is 99% of the case in a separate file/folder than where your models live anyway. It always felt weird to configure the context type as part of your backing/root types, which are arguably two different things.

+1

Weakky: > I thought this was only an internal type (TypegenInfo vs TypegenAutoConfigOptions which doesn't have this), so it didn't matter.

Oh, sorry, you're right I didn't look properly!

jasonkuhrt: Let's take this up in follow up PRs

PR #624, opened 2020-11-15
chore(docs): fix little typo in 021-abstract-types.mdx

fcanela:
> It is based on supplying at `__tpename` field 

Has been corrected to

> It is based on supplying at `__typename` field


fcanela:
I have just noticed that you are using commitizen as standard. Should I replicate the changes with that format?

Sorry, it is the first time I create a commit via the GH interface.

jasonkuhrt (approved):
Thanks!

PR #627, opened 2020-11-16
refactor: remove namespace, separate internal types

tgriesser:
No material changes, just moving some types around. 

I noticed that `AbstractTypes` was a `namespace`. I've generally tried to avoid those (and `enum`'s) in my own work now that there's a way to express them with ES6 `import * as`, so this moves them to `typegenAbstractTypes.ts` and does exactly that, re-exporting as `AbstractTypes` in `_types.ts`

Also adds a `typeHelpersInternal.ts` to contain types we don't consider to be part of the public interface - not that we were ever exporting these anyway, other than on `core.` - but it seemed like it might be good to further mark these as internal.


Weakky (approved):
Great change 👍

jasonkuhrt (approved):
LGTM!

PR #628, opened 2020-11-17
refactor: Simplify typegen with keyof and intersection types

tgriesser:
Was working on re-adding `t.modify` on `objectType` with proper type support to change the resolver for interface fields, and in doing so was looking for access to all the "interface" fields for a given type. 

Realized this was going to require some changes to the types, and while looking at them, realized we can simplify the type generation considerably, by emitting an interface per graphql type, and union'ing them / using `keyof`.


Weakky (approved):
One could argue that things were slightly more explicit before, especially for TS beginners/JS users. I personally don't see it as a blocker though, it def. makes the typegen file cleaner.

One thing I hope is that this doesn't make the type errors any more complicated (due to the intersections etc). Don't see any reason why it should be the case, but who knows what TS does sometimes 😫

tgriesser:
> One thing I hope is that this doesn't make the type errors any more complicated (due to the intersections etc). Don't see any reason why it should be the case, but who knows what TS does sometimes 😫

If it does, I suppose we can revert / reverse the behavior where 
```
export type NexusGenInterfaces = Pick<NexusGenRoot, InterfaceNames>
```

but hopefully we don't have to

PR #634, opened 2020-11-18
feat: add t.modify API for modifying inherited fields

tgriesser:
Adds `t.modify` API for modifying inherited types. 

Use cases:

- If you have an `interface` type, but want to override the resolver or description for the field per-`objectType`
- If an inherited field's value is abstract, and you want to narrow it to a subtype which conforms to the inherited type
- If you want to add additional optional arguments for an inherited field, as is [permitted by the spec](http://spec.graphql.org/draft/#sel-JAHZhCHCDEHDAAACD8ChyL)

The types should validate that only fields inherited from interfaces can be modified, and the `type` can only be changed if the result of that field is an abstract type which implements the existing inherited return type.


tests/modify.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+              type: 'Node',
+            })
+          },
+          resolveType: (o) => o.__typename,
+        }),
+        objectType({
+          name: 'User',
+          definition(t) {
+            t.implements('Node')
+            t.modify('id', {
```
jasonkuhrt, line 37: Why not just support allowing field modifications by simply declaring them. So for example here it could be `t.id(...)`.

Put another way, what's the advantage of having two different APIs for adding fields to a type versus modifying inherited fields? Instead of having one API where adding fields are treated as a modification automatically if a same-named field has been inherited?

tgriesser: > Why not just support allowing field modifications by simply declaring them. So for example here it could be t.id(...).

That's how it works currently, if you define a field, you just use the existing APIs and it overwrites it. 

It's inconvenient though in situations where you have a type with `args` and you want to just change the resolver. Do you need to re-supply the args for that field, or are they merged? If they're merged, how do you know where they're coming from. 

I suppose we could make it so if the field existed already in a parent, you'd just merge them, but for `t.field`, it gets more confusing if you conditionally had to supply `type` dependent on whether or not it is inherited.

I think a `t.modify` API clearly communicates the intent, and restricts the surface area of what you can modify on a field (e.g. can't change the deprecation, or the type unless the inherited field type is an interface).

jasonkuhrt: > I think a t.modify API clearly communicates the intent, and restricts the surface area of what you can modify on a field (e.g. can't change the deprecation, or the type unless the inherited field type is an interface).

Yeah I think this is great. Both the aspect of intent clarification when reading code but also the type safety/clarity. For example I presume that `t.modify(<ctrl+tab>` would show the autocomplete for only inherited fields. That's great!

tests/modify.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+          }
+        }
+      `
+    )
+    expect(result.data?.user.id).toEqual('User:1')
+    expect(result.data?.throws.id).toEqual(null)
+    expect(result.errors?.[0].message).toEqual('Abstract')
+  })
+
+  it('can add additional field args to an interface field', async () => {
```
jasonkuhrt, line 130: Do we have a test to show that it can modify an existing field arg to an interface field? A modification for example being a change of arg type or arg description.

jasonkuhrt (approved):
LGTM

src/definitions/interfaceType.ts
```diff
@@ -1,10 +1,32 @@
...
+  /**
+   * The description to annotate the GraphQL SDL
+   */
+  description?: string | null
+  /**
+   * The resolve method we should be resolving the field with
+   */
+  resolve?: FieldResolver<TypeName, FieldName>
+  /**
+   * You are allowed to add non-required args when modifying a field
```
jasonkuhrt, line 20: Ah I think this answers me previous question, only arg additions are allowed?

Come to think of it this makes sense because otherwise it would violate the interface contract in GraphQL :)

tests/modify.spec.ts
jasonkuhrt: See other comment, no its not supported since it would violate the interface implementation.

I don't think we can enforce this statically though can we? E.g. blacklist fields. Instead we'll have to do runtime checks won't we?

src/definitions/interfaceType.ts
tgriesser: Yeah, only arg additions. I didn't think it mattered much to change the contract of existing args.

tgriesser:
Not sure what's up with the failing tests, but it seems like something related to these changes. Will look into that.

src/definitions/interfaceType.ts
tgriesser: I believe graphql-js will give an error in the schema validation. I wasn't too worried about adding logic for it on our end because it's such an edge case.

PR #635, opened 2020-11-18
[WIP] Plugin to bring back legacy nullable/list API

Weakky:
## Overview

This PR adds a builtin plugin called `legacyNullListApiPlugin`, which brings back the old `nullable|list` API for a smooth upgrade path to the new syntax.

## Open questions

- The old API used to allow `required: boolean` _only_ on scalar shorthand fields. Unfortunately, the plugin API doesn't enable us to differentiate scalar shorthands from regular fields. Therefore, should we:
   - Bring `required: boolean` also on `t.field()`
   - Remove `required: boolean` entirely (thus making a breaking change even with the plugin)
   - Or enhance the plugin API to make the diffenriate scalar shorthand from regular fields?

## TODO:

- [x] Make nullable/list working for input types
- [x] Make nullable/list working for arg types
- [x] Add tests for input/arg types too


tgriesser:
Mentioned this in Slack, but this would be super cool if it were implemented as a plugin, so you could opt-in to the old API if you wanted, and it'd remain type-safe... you could remove when you're ready (or just keep it if you like it), maybe with an option to warn or not. 

Then we could just check if we see `list` / `nullable` accessed and point them to the plugin!

tgriesser:
Somewhat related, I think the SDL converter will probably need to be updated re these changes as well.

Weakky:
> Somewhat related, I think the SDL converter will probably need to be updated around the latest API changes.
Yeah this was mentioned in the TODO of the other PR, I'll make sure to update it in a follow-up PR too!

tgriesser (commented):
Looks pretty good so far. Currently working on some edits to the plugin APIs to simplify a few things. I think it's helping me better understand all the null/list changes. 

If you don't mind holding off merging, I think I'll have a PR to this branch with a few edits tomorrow or this weekend.

tgriesser:
See #655

tgriesser:
Landed in #655

PR #639, opened 2020-11-19
chore: Refine exports for plugins

tgriesser:
Noticed a bunch of things were showing up on the top-level API because of `export *` for various plugins.

We want to group these things under namespaces and keep the top level API clean when we can.


jasonkuhrt (approved):
LGTM!

PR #643, opened 2020-11-20
refactor: move wrapping helper fns to definitions/wrapping.ts

tgriesser:
Minor, noticed these maybe were better placed in `wrapping.ts` than `utils.ts`


jasonkuhrt (commented):
LGTM!

PR #649, opened 2020-11-20
deps: update to latest

### LINKED ISSUE #580: Raise minimum required TS version to 4.1
TS 4.1 will have a lot of new and interesting power for complex type expressions. We should lock in that requirement for Nexus before hitting 1.0. This way we can start to roll out features after without requiring to hit this breaking change _after_ 1.0 so soon.

There might not be any immediate end-user benefits by raising the TS bar to 4.1 however improved internal typings and typegen will be indirectly important to project health and momentum.

jasonkuhrt:
closes #580

Basically we now say they should use TS 4.1. And in the future we'll force that without introducing a breaking change... I guess. 😕 

It seems pedantic but we could apply a TS 4.1 feature somewhere just to force the upgrade and avoid confusion later. E.g.:

```ts
type Foobar = 'hello'
type ForceTS41 = `${Foobar} World`
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,13 +69,13 @@
     ]
   },
   "dependencies": {
-    "iterall": "^1.2.2",
-    "tslib": "^1.9.3"
+    "iterall": "^1.3.0",
+    "tslib": "^2.0.3"
   },
   "devDependencies": {
     "@types/graphql-iso-date": "^3.4.0",
     "@types/graphql-relay": "^0.4.11",
-    "@types/jest": "^25.2.3",
+    "@types/jest": "^26.0.15",
     "@types/node": "^10.12.2",
     "@types/prettier": "^1.18.3",
     "@typescript-eslint/eslint-plugin": "2.7.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -69,13 +69,13 @@
     ]
   },
   "dependencies": {
-    "iterall": "^1.2.2",
-    "tslib": "^1.9.3"
+    "iterall": "^1.3.0",
+    "tslib": "^2.0.3"
```
jasonkuhrt, line 73: Note https://github.com/microsoft/tslib/releases/tag/2.0.0

Requires TS 3.9+

But should be fine for us here since https://github.com/graphql-nexus/schema/issues/580

tgriesser: Prettier needs to be bumped to 2.2 if you want to use any of the new template string features FYI, not that needs to happen here since we're not using any yet (and should probably hold off unless there's a compelling need).

tgriesser: Does tslib need to change to use 4.1?

jasonkuhrt: > Does tslib need to change to use 4.1?

Not that I know of, but I figured we should update it as a matter of housekeeping.

PR #650, opened 2020-11-20
feat: add t.nonNull and t.nullable shorthands

Weakky:
## Overview

- Adds `t.nonNull` and `t.nullable` shorthands in addition to the existing `t.list` shorthand
- Shorthands cannot be used in combination with type wrapers (list(), nonNull(), nullable())

## TODO

- [ ] Update examples with new shorthands where it makes sense
- [ ] Update docs


src/builder.ts
```diff
@@ -1385,24 +1403,30 @@ export class SchemaBuilder {
...
       if (typeof val === 'string') {
         return this.addDynamicScalar(methodName, val, block)
       }
       // @ts-ignore
       block[methodName] = (...args: any[]) => {
-        const config = isList ? [args[0], { list: isList, ...args[1] }] : args
+        const wrappedType =
+          args[1]?.type !== undefined ? wrapAsNexusType(args[1].type, wrapping, false) : undefined
+        const config = wrappedType !== undefined ? [args[0], { ...args[1], type: wrappedType }] : args
+
```
Weakky, line 1429: Not so sure what to do here now...

src/builder.ts
```diff
@@ -1413,14 +1437,21 @@ export class SchemaBuilder {
...
```
Weakky, line 1454: Same here, not so sure what to do now @tgriesser

Weakky: Makes me think this is actually wrong and need to be updated anyway, the `nonNullDefault` is always put to false, but my guts tell me this should be handled differently anyway

src/definitions/definitionBlocks.ts
```diff
@@ -158,16 +187,16 @@ export class OutputDefinitionBlock<TypeName extends string> {
...
-    if (this.isList) {
-      if (isNexusListTypeDef(config.type)) {
-        this.typeBuilder.warn(
-          `It looks like you chained .list and used list() for ${config.name}. ` +
-            'You should only do one or the other'
+    if (isNexusWrappingType(config.type)) {
+      if (this.wrapping.length > 1) {
+        throw new Error(
+          'Cannot use t.list|nonNull|nullable shorthands and list()|nonNull()|null() at the same time'
         )
```
Weakky, line 194: If you want a better error message, lemme know. I'm bad at these.

src/definitions/definitionBlocks.ts
```diff
@@ -192,16 +221,40 @@ export interface InputDefinitionBlock<TypeName extends string> extends NexusGenC
 
 export class InputDefinitionBlock<TypeName extends string> {
   readonly typeName: string
-  constructor(protected typeBuilder: InputDefinitionBuilder, protected isList = false) {
+  constructor(
+    protected typeBuilder: InputDefinitionBuilder,
+    protected nonNullDefault: boolean,
```
tgriesser, line 226: Why do we need to pass `notNullDefault` through here?

src/definitions/wrapping.ts
```diff
@@ -164,61 +164,32 @@ export function isNexusPlugin(obj: any): obj is NexusPlugin {
   return isNexusStruct(obj) && obj[NexusWrappedSymbol] === NexusTypes.Plugin
 }
 
-type NexusWrapKind = { type: 'List'; nullable: boolean } | { type: 'NamedType'; nullable: boolean }
+export type NexusWrapKind = 'NonNull' | 'Null' | 'List' | 'WrappedDef'
```
tgriesser, line 167: What is `WrappedDef` for?

src/builder.ts
tgriesser: What do you mean?

src/definitions/definitionBlocks.ts
Weakky: Because we need the nonNullDefault to be passed to the wrapping function.. That function needs the nonNull Default to properly add the `nonNull` if it's nonNull by default

src/definitions/wrapping.ts
Weakky: list('String') -> ['WrappedDef', 'List]. We need the WrappedDef to be able to apply the nonNull() if nonNullDefault is true

src/definitions/definitionBlocks.ts
tgriesser: Could we source that later in the building phase when the wrapping occurs, or did something change here that I'm missing compared to the previous behavior?

src/definitions/wrapping.ts
tgriesser: I see, makes sense.

src/definitions/definitionBlocks.ts
Weakky: So what's technically changed is that previously we were just setting a config (list:true) vs. now it's the actual type that we need to change.

Couple solutions:

- We get the `wrapping` out of the definition block thanks to the typeBuilder, and we apply wrap the type outside of it

- It's a bit dirty but we actually don't need to apply the proper nonNullDefault at this stage, since we'll need to unwrap/rewrap with graphql types later anyway, and with this PR we're now applying the nonNullDefaults when wrapping.

tgriesser (changes requested):
Working on some changes to simplify a few things in the PR, and extend support to non-nexus types. Will open a PR against this branch

tgriesser:
See #655

tgriesser:
Landed in #655

PR #652, opened 2020-11-20
feat(config): SDL file generation by default in development

### LINKED ISSUE #632: Enable SDL output by default
I think by default the SDL output should be enabled and should output to the first ancestor directory where a `package.json` is found. The file name should default to `schema.graphql`.

jasonkuhrt:
closes #632

BREAKING CHANGE:

SDL file generation will be enabled by default in development now.

If you were enabling it manually before and outputting to project root at `schema.graphql` You can probably now just rely on the default.

If you were relying on the default to disable SDL file generation before then now you need to pass an explicit false:

```ts
makeSchema({
  outputs: {
    schema: false
  }
})
```

### DIFF
--- a/src/typegenUtils.ts
+++ b/src/typegenUtils.ts
@@ -1,11 +1,10 @@
+import * as Path from 'path'
 import { BuilderConfigInput } from './builder'
 import { TypegenMetadataConfig } from './typegenMetadata'
-import { assertAbsolutePath, getOwnPackage } from './utils'
+import { assertAbsolutePath, getOwnPackage, isProductionStage } from './utils'
 
 /**
  * Normalizes the builder config into the config we need for typegen
- *
- * @param config {BuilderConfigInput}
  */
 export function resolveTypegenConfig(config: BuilderConfigInput): TypegenMetadataConfig {
   const {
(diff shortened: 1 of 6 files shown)


src/typegenUtils.ts
```diff
@@ -1,11 +1,10 @@
+import * as Path from 'path'
```
tgriesser, line 1:
```suggestion
import * as path from 'path'
```

src/typegenUtils.ts
```diff
@@ -14,14 +13,27 @@ export function resolveTypegenConfig(config: BuilderConfigInput): TypegenMetadat
     ...rest
   } = config
 
-  let typegenPath: string | false = false
-  let schemaPath: string | false = false
-  if (outputs && typeof outputs === 'object') {
-    if (typeof outputs.schema === 'string') {
-      schemaPath = assertAbsolutePath(outputs.schema, 'outputs.schema')
+  const defaultSDLFilePath = Path.join(process.cwd(), 'schema.graphql')
```
tgriesser, line 16:
```suggestion
  const defaultSDLFilePath = path.join(process.cwd(), 'schema.graphql')
```

tgriesser (changes requested):

PR #655, opened 2020-11-22
feat: .nonNull / .nullable chaining, additional cleanup

tgriesser:
- Fixes SDL conversion
- Defer wrapping until types are built
- Supports GraphQL types anywhere that a type can be used, wrapped, etc.
- Split out the wrapping, w/ `finalizeWrapping` helper which adds in the `NonNull` to the list as needed
- Make `nonNull(nonNull(T))` & `nullable(nullable(T))` no-op for simplicity
- Change: `nullable(nonNull(T))` will undo the `nonNull`
- Properly type-check's `subscriptionType` by passing `Event` generic properly
- Adds `declarativeWrappingPlugin` for previous `nullable` / `list` behavior


docs/content/015-api/061-list-nonNull.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: list / nonNull / nullable
+title: list / nonNull
```
tgriesser, line 2: Since we default to `nonNullDefault: false`, I wanted to avoid advertising `nullable` in the top-level API. It's a bit of an antipattern.

tests/__snapshots__/null-list.spec.ts.snap
```diff
@@ -608,333 +608,333 @@ exports[`wrapping for args; nonNullDefault = true; wrapped arg def nullable(list
...
+exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list.list.nullable.field({type: String}) 1`] = `"[[String]!]!"`;
 
-exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list(nonNull(String)) 1`] = `"[String!]!"`;
+exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list.nonNull.field({type: String}) 1`] = `"[String!]!"`;
 
-exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list(nonNull(list(String))) 1`] = `"[[String!]!]!"`;
+exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list.nonNull.list.field({type: String}) 1`] = `"[[String!]!]!"`;
 
-exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list(nonNull(nonNull(String))) 1`] = `"Error: Cannot wrap a nonNull() in a nonNull()"`;
+exports[`wrapping for input types with chained API; nonNullDefault = false; string ref list.nonNull.nonNull.field({type: String}) 1`] = `"Error: Expected String! to be a GraphQL nullable type."`;
```
Weakky, line 719: Something looks wrong here. This shouldn't happen this we're making `nonNull.nonNull` a no-op now

Weakky: Investigated the reason. That's because we're not preventing multiple `NonNull` or `Null` to be unshifted to the `wrapping` list in the definition blocks. I have a simple fix but I'm wondering whether you kept it that way intentionally. I'll let you decide.

src/builder.ts
```diff
@@ -1337,6 +1344,19 @@ export class SchemaBuilder {
...
+      if ('list' in def || 'nullable' in def || (def.configFor === 'arg' && 'required' in def)) {
+        const d = def as NexusFinalArgConfig | NexusOutputFieldDef | NexusInputFieldDef
+        let location =
+          d.configFor === 'arg'
+            ? `'${d.parentType}.${d.fieldName}' field's '${d.argName}' argument`
+            : `'${d.parentType}.${d.type}' field`
+        throw new Error(messages.removedDeclarativeWrapping(location))
+      }
+    }
+  }
```
Weakky, line 1358: Yeah, forgot about that one, that's a great addition!

Weakky (approved):
Great job, LGTM 👍

src/definitions/wrapping.ts
```diff
@@ -164,127 +172,158 @@ export function isNexusPlugin(obj: any): obj is NexusPlugin {
...
+      if (nonNullDefault && (next === 'List' || !next)) {
+        finalChain.push('NonNull')
+      }
     }
-
-    return type
-  }, baseType)
+  }
+  return finalChain
 }
```
Weakky, line 329: Errrr. I swear I tried the very same approach cause the other approach was terribly ugly but didn't manage to have it working 😫. What I wasn't doing properly when I tried was to keep reading the wrapping by pairs when inserting back the `NonNull`'s 🤦‍♂️

src/plugins/connectionPlugin.ts
```diff
@@ -145,11 +143,6 @@ export type NodeValue<TypeName extends string = any, FieldName extends string =
 
 export type ConnectionFieldConfig<TypeName extends string = any, FieldName extends string = any> = {
   type: GetGen<'allOutputTypes', string> | AllNexusNamedOutputTypeDefs
-  /**
-   * Whether the connection field can be null
-   * @default (depends on whether nullability is configured in type or schema)
-   */
-  nullable?: boolean
```
Weakky, line 152: Before going into API design, I think you need to update the TS type of the `type` property to also accept `NexusNonNullDef<any> | GraphQLNonNull | ..`

Now, about that, we initially intentionally kept the `nullable: boolean` API on the connection plugin, because it felt wrong to express the nullability of the generated `Connection` type with wrappers. Example:

```ts
t.connectionField('booleanConnection', {
  type: nonNull('Boolean')
})
```
Will generate the following field:
```graphql
booleanConnection: BooleanConnection!
```

Using `nonNull('Boolean')` to generate `BooleanConnection!` feels a bit far-fetched, but I'll let you decide what you prefer

src/definitions/wrapping.ts
tgriesser: Haha. To be fair this took me like two hours of staring at it and trying to figure out what it should be doing. Wouldn’t have been able to get it without the existing test suites you added.

src/plugins/connectionPlugin.ts
tgriesser: Great point, I’m good with keeping the nullable api here! However, `t.nonNull.connectionField` should work too, right?

tests/__snapshots__/null-list.spec.ts.snap
tgriesser: Hmm. Yeah it’s not intended you should ever do this, but yeah I missed this case, because the wrapping array gets applied at the end. What fix were you thinking?

Weakky: That was my fix 👇 

image

src/plugins/connectionPlugin.ts
Weakky: Yes, fully agreed! `t.nonNull` should just set `nullable: false`

PR #663, opened 2020-11-24
feat: Simplified upgrade path w/ better warning messages

tgriesser:
I just spent about 2 hours trying to resolve the issues I had between 0.16 and now on several of our schemas, and still couldn't get things fully up to speed. Some of it was due to things like #661, but I figured if I was having that hard of a time with hundreds of types spread across a fairly large repo, and I created the library, we could use a better story around migrating.

The main thing is that if we intend to make breaking changes we should should warn and remap the behavior, rather than throw, whenever we can - If the schema doesn't build, the types aren't up-to-date w/ the changes, which isn't a great experience.

- Installs the `declarativeWrappingPlugin({ shouldWarn: true })` if you haven't specified it. 
  - Adding `[declarativeWrappingPlugin()]` will remove the warnings
  - Adding `[declarativeWrappingPlugin({ disable: true })]` will remove the typings, and make using these properties throw an error rather than a `console.error`.
  - In 1.0 we'll change this to `declarativeWrappingPlugin({ disable: true })`
  - In 1.1 we'll remove it
- Flips to `resolveType` as the default strategy for new abstract changes
  - It's more familiar to folks coming from Apollo
  - It's the migration path anyone who has used Nexus in the last 1.5 years would be more familiar with
  - It's better for getting started / smaller schemas, or if you are more bought into a monolithic approach to development
- Keeps `t.resolveType` supported at runtime with a warning, making it a protected member on the class
  - It was simple to add, and there's no real harm to doing this - it helps the schema compile, as I mentioned
  - At 1.0 we'll throw an error here and not remap
  - At 1.1 we'll remove completely


src/messages.ts
```diff
@@ -1,21 +1,40 @@
 export const messages = {
   /* istanbul ignore next */
   removedResolveType: (location: string) => `\
-The .resolveType used in the ${location} has been removed in favor of a more robust approach to handling abstract types.
+The .resolveType used in the ${location} has been moved to a property on the type definition object, 
+as part of of a more robust approach to handling abstract types.
 
 Visit https://nexusjs.org/docs/guides/abstract-types for an explanation and upgrade info.
```
jasonkuhrt, line 7:
```suggestion
Visit https://nxs.li/guides/abstract-types for an explanation and upgrade info.
```

jasonkuhrt (approved):
LGTM!

PR #677, opened 2020-11-26
feat: support graphql@^14.5

tgriesser:
The only major change between `14.5` and `15` was the addition of interfaces implementing interfaces. 

Because migrating from 14 to 15 can sometimes require quite a bit of work due to transitive dependencies, I'd say we should be able to support `graphql@14` for just a bit longer, with some added shims.


jasonkuhrt (approved):
LGTM

PR #682, opened 2020-11-27
fix: #618 Add inputFieldDefTypes for declarativeWrappingPlugin

### LINKED ISSUE #618: Make it easier to evolve the schema in a backward compatible way
I have been experimenting with https://github.com/ef-eng/graphql-query-rewriter to allow schema breaking changes without breaking existing clients (a la stripe with the version in the header).
It would be nice if it was somewhat integrated in nexus with some type safety. Not sure how that would look like yet.

tgriesser:
Fixes #618

### DIFF
--- a/src/definitions/definitionBlocks.ts
+++ b/src/definitions/definitionBlocks.ts
@@ -31,6 +31,15 @@ export type CommonInputFieldConfig<TypeName extends string, FieldName extends st
   default?: GetGen3<'inputTypes', TypeName, FieldName>
 } & NexusGenPluginFieldConfig<TypeName, FieldName>
 
+/**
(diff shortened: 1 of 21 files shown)


Weakky (approved):
Looks all good to me, especially with that integration test 💯

Sytten:
Is there some documentation yet on that feature, it sounds awesome!

tgriesser:
Not really a feature, just allows for defining typings specific to input object type fields similar to output object fields, which was needed for the `declarativeWrappingPlugin` to work without type errors.

tgriesser:
I can add some docs for it though

Sytten:
Yeah would be nice, I would like to write a plugin similar to the query rewriter for nexus if possible to have backward compatible schemas.

Issue #673, opened 2020-11-25
Allow extendType & extendInputType to extend a GraphQL Type


jasonkuhrt:
```ts
import { GraphQLObjectType } from 'graphql'
import { extendType } from '@nexus/schema' 

const A = new GraphQLObjectType()

const B = extendType({
	type: 'A',
	// ...
})

makeSchema({
	types: [A,B]
})
```

jasonkuhrt:
@tgriesser WDYT?

tgriesser:
Does this not already work?

jasonkuhrt:
@Weakky has more context than me, apparently not.

tgriesser:
I'm surprised it doesn't, this is the whole reason we walk & unwrap the types as `.toConfig` - specifically so `extendType` will work - guess it'd be good to start from a failing test case.

tgriesser:
Ah, yep this is a bug.

tgriesser:
This will be fixed by #983

PR #684, opened 2020-11-27
feat: allow arbitrary extensions on nexus types

### LINKED ISSUE #683: How to set extensions on a type?
I found out that it is possible to extend the extensions on a field using the queryComplexityPlugin as a blueprint.
But now I would like to do the same for a type as well.

For my particular use case I am trying to use JoinMonster with @nexus/schema. For this to work JoinMonster (starting with version 3.0.0, see https://github.com/join-monster/join-monster/blob/master/docs/CHANGELOG.md#v300) requires to use the GraphQL extensions property introduced in GraphQL v15.

As a simple solution I tried to extend the queryComplexityPlugin example and in ```onCreateFieldResolver(config)``` run the following code to update the type extensions config:

```
const joinMonsterConfigOnType =
    config.parentTypeConfig.extensions?.nexus?.config.joinMonster;

if (joinMonsterConfigOnType != null) {
    config.parentTypeConfig.extensions = {
        ...config.parentTypeConfig.extensions,
        joinMonster: joinMonsterConfigOnType,
    } as any;
}
```

But it looks like Nexus is overwriting the type's extensions later on with its own. Could be this location here: https://github.com/graphql-nexus/schema/blob/develop/src/builder.ts#L963

Any help would be much appreciated.

tgriesser:
Closes #683


jasonkuhrt:
pr release made:

```json5
 { distTag: 'pr.684', version: '0.0.0-pr.684.1.187f285' }
```

benjixx:
I just discovered an edge case which seems to be not working yet.

I have an interface A and some objectType B that implements interface A. I would like to modify the implemented field that is part of interface A (via modify method) within objectType B and also set the extension for the field there.

jasonkuhrt (approved):
LGTM, maybe use the feature in the integration tests to "cover" this?

tgriesser:
> I have an interface A and some objectType B that implements interface A. I would like to modify the implemented field that is part of interface A (via modify method) within objectType B and also set the extension for the field there.

Cool, will look into supporting this

jasonkuhrt:
@benjixx added a failing test for that case.

tests/interfaceType.spec.ts
```diff
@@ -196,3 +196,49 @@ describe('interfaceType', () => {
...
+    ],
+    outputs: false,
+    shouldGenerateArtifacts: false,
+  })
+  const A = schema.getType('A') as GraphQLObjectType
+  expect(A.getFields().foo.extensions).toMatchInlineSnapshot(`
+    Object {
+      "deeply": Object {
+        "foo1": true,
+        "foo2": true,
```
jasonkuhrt, line 240: @tgriesser do you agree extension modifications should deeply merge?

tests/integrations/kitchenSink/__typegen.ts
```diff
@@ -80,10 +80,7 @@ export interface NexusGenObjects {
     // root type
     hello?: string | null // String
   }
-  OfI2: {
-    // root type
-    hello?: string | null // String
```
jasonkuhrt, line 85: @tgriesser Looks like a bug to me. If user uses `t.modify` only to augment `extensions` then why should the default backing types be altered?

tests/interfaceType.spec.ts
tgriesser: I think they should probably stay independent until we have a use case for them to be merged.

tests/integrations/kitchenSink/__typegen.ts
tgriesser: Yep, this is all pretty messy right now. Got it working temporarily, but will try and refactor this as a part of fixing #673

PR #691, opened 2020-11-30
chore(docs): add nexus next.js adoption guide

### LINKED ISSUE #543: Create an adoption guide for Next.js users

chenkie:
closes #543


docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+Next.js comes with a feature called [API Routes](https://nextjs.org/docs/api-routes/introduction). With API Routes, we can easily create dedicated endpoints to use in our Next.js applications. These endpoints can then be deployed as serverless functions.
+
+In this guide, we'll see how to work with Next.js API Routes to serve a Nexus Schema GraphQL API and then consume it from the front end.
+
+## Getting Started
+
+Let's start by creating a new Next.js application. If you already have a Next.js app, you can ignore this step.
+
+```bash
+npx create-next-app
```
jasonkuhrt, line 16: How about `npm init next-app` or `yarn create next-app`?

https://create-react-app.dev/docs/getting-started/#npm

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+
+```bash
+npm run dev
+```
+
+For more information on using TypeScript in Next.js, check out [the docs](https://nextjs.org/docs/basic-features/typescript).
+
+With the Next.js project in place, we now need to install the other dependencies we'll need.
+
+In addition to Nexus Schema, we'll need a server to run the GraphQL API. Nexus Schema works with a range of GraphQL servers. We'll use Apollo Server and opt for the `apollo-server-micro` variety as it is well-suited to serverless environments.
```
jasonkuhrt, line 42: We're moving back toward just saying "Nexus". Package will be renamed to `nexus` in 1.0.

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+touch index.ts
+```
+
+In the `index.ts` file, bring in `makeSchema` as well as the types we just created in `Query.ts`.
+
+```ts
+// schema/index.ts
+
+import { makeSchema } from '@nexus/schema'
+import { Framework, Query } from './Query'
```
jasonkuhrt, line 182:
```suggestion
import * as QueryTypes from './Query'
```

Would it be simpler like this so that there is a direct passthrough established between the module exports and the makeSchema config?

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+const schema = makeSchema({
+  types: [Framework, Query],
+})
+
+export default schema
+```
+
+Now in the `graphql.ts` for our API route, we can simple pull in the constructed schema and serve it with Apollo Server.
+
+```ts
```
jasonkuhrt, line 193: Maybe use a diff here?

` ```ts diff`

Also you can use tabs to have a "diff" tab and a "copy-able" tab.

For reference see tutorial source code here https://github.com/graphql-nexus/schema/blob/develop/docs/content/010-getting-started/03-tutorial/03-chapter-2-writing-your-first-schema.mdx#L74.

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+})
+```
+
+Using `__dirname` in this fashion will not work in a Next.js project. Next.js takes control of `__dirname` and overrides it with special behavior which interferes with our ability to use it as we normally would.
+
+The solution is to use `path.join` and `process.cwd` together to describe where to place the generated files.
+
+In the `schema/index.ts` file, configure `makeSchema` to output the generated SDL and typings using `path.join` and `process.cwd`.
+
+```ts diff
```
jasonkuhrt, line 300: Ah, you already know about this :)

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+}
+```
+
+When the page renders, we'll see a list of the frameworks.
+
+![a next.js page rendered with a list of frameworks](/assets/next-nexus-3.png)
+
+**Note:** Using `graphql-request` to make a server-side call for data from our GraphQL API is good for initially fetching the data but it won't do much for us if we want to have client-side GraphQL features such as caching. If we want additional features like those provided by Apollo Client, it won't work to mix the data retrieved in the `getServerSideProps` function with data managed on the client by Apollo.
+
+## Schema and Type Generation
```
jasonkuhrt, line 280:
```suggestion
## SDL and Type Generation
```

Avoid ambiguity of word "schema"?

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+
+A GraphQL SDL file and Nexus typings can automatically be generated by configuring the `outputs` option in `makeSchema`. In a typical project, we can tell Nexus where to place these files by using `__dirname` to define a path. That might looks something like this:
```
jasonkuhrt, line 282: Since these files will be generated by default soon. So something like this?

```suggestion
Nexus generates a GraphQL SDL file and TypeScript typings by default. You can customize their location by configuring `outputs` option in `makeSchema`. In a typical project, we can tell Nexus where to place these files by using `__dirname` to define a path. That might looks something like this:
```

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
++      process.cwd(),
++      'generated',
++      'schema.graphql'
++    )
+  }
+});
+
+export default schema;
+```
+
```
jasonkuhrt, line 323: How would the build command work? Before `tsc` runs Nexus needs the typings file to be present. Generally we suggest an npm script like `"build": "ts-node --transpileOnly path/to/schema"`. How should it work in this case? Based on what you've written I think we can do this:

```json
{
  "scripts": {
    "build:nexus-typegen": "ts-node --transpileOnly schema",
    "build": "yarn build:nexus-typegen && next build"
  }
}
```

?

jasonkuhrt (commented):
Nice!!

I think it would be nice to add a link to a Next.js Codesandbox at the end (or beginning). WDTY?

I've added you to the Nexus workspace https://codesandbox.io/dashboard/home?workspace=898d6824-642a-4478-a544-52cd5361de7e.

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
+---
+title: Nexus With Next.js
```
jasonkuhrt, line 2: I think the title should match with the pattern of other adoption guides. Either rename this to `Next.js Users` or rename the others to be e.g. `Nexus with Prisma`, `...`. Though I don't know how that's going to work for the `Nexus Framework Users` title... So maybe `Next.js Users` is best & most consistent?

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
...
+            name: 'Svelte',
+          },
+        ]
+      },
+    })
+  },
+})
+```
+
+Let's now create a barrel file in the `schema` directory which will be responsible for importing all the types in the directory and exporting a constructed schema using `makeSchema`.
```
Weakky, line 169: Small note: I don't know what a `barrel file` is. I'm sure the meaning of the sentence is not lost at all anyway, but _maybe_ it's a sign that we should use a simpler word. This _isn't_ a strong suggestion, I'll leave it up to you

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
Weakky: I favor this change too. This is considered a best practice of nexus and should be used as much as possible in our examples/tutorial to teach this pattern

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
jasonkuhrt: Almost said the same thing actually! Also I'm not sure this is a barrel file since it has logic, not just exports.

PR #692, opened 2020-12-01
docs: add sticky table of content

Weakky:
Dunno what I was thinking to put "nested wrapping" as commit name 🤦 

## Notes

- `stickWhenNeeded` helper was taken from `prisma/docs`

### DIFF
--- a/docs/content/014-guides/020-schema.mdx
+++ b/docs/content/014-guides/020-schema.mdx
@@ -1,5 +1,6 @@
 ---
 title: 'Schema'
+tocDepth: 2
 ---
 
 ## Overview
@@ -309,7 +310,7 @@ When you make an input nullable then Nexus will alter its TypeScript type inside
 
 If an arg has been given a default value, then it will be used when the client passes nothing, but since clients can still pass explicit `null`, resolvers must still handle nullability. If this surprises you then you may be interested in [#485](https://github.com/graphql-nexus/nexus/issues/485).
 
-#### Example: Default nullability settings
+### Example: Default nullability settings
 
 
 
@@ -336,7 +337,7 @@ type Query {
 
 
 
-#### Example: Nullability flipped at global level
+### Example: Nullability flipped at global level
 
 ```ts
 makeSchema({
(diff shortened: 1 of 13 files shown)


docs/content/014-guides/020-schema.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: 'Schema'
+tocDepth: 2
```
Weakky, line 3: Allows specifying the TOC depth

docs/src/layouts/articleLayout.tsx
```diff
@@ -8,7 +8,15 @@ import SEO from '../components/seo'
 import { graphql } from 'gatsby'
 import MDXRenderer from 'gatsby-plugin-mdx/mdx-renderer'
 
-type ArticleLayoutProps = ArticleQueryData & RouterProps
+const DEFAULT_TOC_DEPTH = 3
```
Weakky, line 11: Default TOC depth is 3. That's voluntarily a pretty high number so that we have as many details as possible, until it becomes too much and then we can reduce it using the `tocDepth: number` mdx header property

docs/src/utils/stickWhenNeeded.ts
```diff
@@ -0,0 +1,66 @@
...
+      var sidebarBottom = sidebarTop + sidebarHeight
+
+      var heightDelta = Math.abs(windowHeight - sidebarHeight)
+      var scrollDelta = lastScrollTop - scrollTop
+
+      var isScrollingDown = scrollTop > lastScrollTop
+      var isWindowLarger = windowHeight > sidebarHeight
+
+      // Hack: gradually increase the height of the container as we scroll down
+      $sidebar.css('height', `calc(100vh - ${Math.max(0, sidebarTop - scrollTop)}px)`)
```
Weakky, line 29: This is bad but meh, most of it's gonna go away at some point anyway. This sets the proper height of the TOC container relative to the scroll position. It basically goes from 150px (size of the header) to 0px when we reach the sticky position.

PR #704, opened 2020-12-05
docs(chore) : fix typo

redrebel:
Are they generate in alphabetical order?

```
export const Post = objectType({
  name: "Post",
  definition(t) {
    t.int("id");
    t.string("title");
    t.string("body");
    t.boolean("published");
  },
});
```

```
type Post {
  body: String
  id: Int
  published: Boolean
  title: String
}
```


jasonkuhrt (approved):
Ty!

Issue #678, opened 2020-11-26
Testing documentation tutorial results in errors


saphewilliam:
Hey all! 

First of all, I'd like to say that I am very enthusiastic about the future of nexus and prisma. It's been great fun learning these tools the past few weeks!
I've started an attempt to write a test suite for an API I'm working on, following the tutorial on the nexus documentation site. 
https://nexusjs.org/docs/getting-started/tutorial/chapter-testing-your-api

However, two things arose as I was following this step: 
* The pm add command in the article `npm add --save-dev jest @types/jest ts-jest graphql-request` does not mention the `get-port` npm package which is later used in `tests/__helpers.ts`
* After running my tests, jest gives the error message `Jest did not exit one second after the test run has completed. This usually means that there are asynchronous operations that weren't stopped in your tests.` and doesn't terminate. Running with the `--detectOpenHandles` flag hides the warning message but jest still doesn't terminate and shows no extra output.

Maybe a pointer towards the solution of the second issue: when I run a development server of the api on port 4000 and I have a test suite with only one test which I also run on port 4000, the warning is not displayed and jest exits as normal.

The project I'm working on is based on the code in this project from the prisma-examples repo: https://github.com/prisma/prisma-examples/tree/latest/typescript/subscriptions-pubsub. 

I am not experienced with jest (or rather, software testing as a whole) and I would love some help. Is anyone else experiencing the same issue?

shreyas44:
@saphewilliam I'm completely new to testing as well. Were you able to get around or fix this problem?

saphewilliam:
I have not yet, are you experiencing the same issues as me?

I did find a prisma day 2020 video where they explain testing with nexus-plugin-prisma. I have not had the time to test it yet, it might be a bit outdated. https://youtu.be/oFk4rxz_KO8?t=8328 (timestamp included)

Are you willing to test it and report back to this issue if it is still working?

shreyas44:
@saphewilliam Sure! I can look into the video you sent and explore more as well and get back to you. Meanwhile if you come across a solution, do update 🙂 .  

I used Apollo Express server rather than the standalone Apollo server and one solution I tried was closing the express server in the `afterEach` hook but that didn't seem to work either.

jasonkuhrt:
Pass the `--forceExit` flag to jest. I fixed a bunch of things in the tutorial tonight and updated https://github.com/graphql-nexus/tutorial. Hope it helps!

saphewilliam:
[...]

PR #709, opened 2020-12-07
chore(docs): update tutorial testing chapter

shreyas44:
- Use the Prisma Client created in `db.ts` in the API Context
- Disconnect Prisma Client after running each test, allowing Jest to exit. Previously this was an issue as the Prisma Client connection was never closed and prevented Jest from exiting. This issue was fixed in [`a9ca189`](https://github.com/graphql-nexus/tutorial/commit/8bcbe015fa1779770caa6b9dbcce67d000ba5c9c) using a workaround by passing the `--forceExit` flag to Jest.

This PR adds a proper fix for #678

### DIFF
--- a/docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
+++ b/docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
@@ -204,6 +204,7 @@ Then, update your `Context` type so that it uses the PrismaClient created in `./
 ```ts
 // api/context.ts
 import { db } from "./db";
++import { PrismaClient } from "@prisma/client"
 
 export interface Context {
 -  db: Db
@@ -222,6 +223,7 @@ export function createContext(): Context {
 ```ts
 // api/context.ts
 import { db } from "./db";
+import { PrismaClient } from "@prisma/client"
 
 export interface Context {
   db: PrismaClient
--- a/docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
+++ b/docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
@@ -196,24 +196,22 @@ export const db = new PrismaClient()
 </TabbedContent>
 
 
-Then, update your `Context` type so that it uses the PrismaClient
+Then, update your `Context` type so that it uses the PrismaClient created in `./db.ts`
 
 <TabbedContent tabs={['Diff', 'Code']}>
 <tab>
 
 ```ts
 // api/context.ts
--import { Db, db } from "./db";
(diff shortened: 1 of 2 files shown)


docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
```diff
@@ -223,15 +222,16 @@ export function createContext(): Context {
...
+import { PrismaClient } from "@prisma/client"
 
 export interface Context {
   db: PrismaClient
 }
 
 export function createContext(): Context {
   return {
-    db: new PrismaClient(),
+    db,
```
jasonkuhrt, line 234: Can you update this to match https://github.com/graphql-nexus/tutorial/blob/master/api/context.ts

I don't think we need a db modular just to instantiate a Prisma client :)

jasonkuhrt, line 234: Can you update this to match https://github.com/graphql-nexus/tutorial/blob/master/api/context.ts

I don't think we need a db modular just to instantiate a Prisma client :)

jasonkuhrt (changes requested):

jasonkuhrt:
Awesome! Just one thing

docs/content/010-getting-started/03-tutorial/06-chapter-5-persisting-data-via-prisma.mdx
shreyas44: We don't need db to modular, but we do need to store the instance of the Prisma Client in either an object or variable which is exported so that it can be accessed in `tests/__helpers.ts`. The same Prisma Client used in the context of Nexus should be accessible in `__helpers.ts` as a new connection is created for every instance of Prisma Client. 

There are two options, one is keep it as I've updated it.

Another is, to export an object from 'context.ts' and disconnect the connection of that
```ts
export const context: Context = {
  db: new PrismaClient()
}
```
and in `__helpers.ts`
```ts
import { context } from '../api/context.ts'

... 

function graphqlTestContext() {
  ...
  
  return {
    async before() {
      ...
      serverInstance.server.on('close', async () => {
        await context.db.$disconnect()
      })
    },
    ...
  }
}
```

jasonkuhrt: Yep good point. then a singleton db module is a simple solution indeed. If you're inclined update the tutorial repo otherwise we'll try to do it soon.

shreyas44: Already done graphql-nexus/tutorial#13 🙂 !

Issue #695, opened 2020-12-02
Documentation on makeSchema/typegenAutoConfig


JulianaAmoasei:
I'm looking for documentation on the use of the `sources` property on `typegenAutoConfig`, but ended up with more doubts... The [related page](https://nexusjs.org/docs/api/make-schema#shouldgenerateartifacts-outputs-typegenautoconfig) in the docs seems to be incomplete, as some sentences kind of end abruptly.

In the code below, I used the module `@prisma/client` as source for type generation, but for some reason the API still works either I declare `sources` or not. I'm not sure I understood how this property works and how to use it...

I'd be grateful if someone could clarify it for me!

```
const schema = makeSchema({
  types: [Query, Mutation, Post, User, Review],
  plugins: [nexusPrisma({ experimentalCRUD: true })],
  outputs: {
    schema: path.join(__dirname, 'schema.graphql'),
    typegen: path.join(__dirname, '../prisma/generated', 'nexus.ts'),
  },
  typegenAutoConfig: {
    sources: [
      {
        source: '@prisma/client',
        alias: 'prisma',
      },
    ],
  },
})
```

PR #711, opened 2020-12-07
docs: move source type section into its own guide

Weakky:
closes https://github.com/graphql-nexus/schema/issues/695


docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+
+![](/assets/source-type-diagram-1.png)
+
+When a field's type is an object, then the field's resolver returns a source type. Concretely this might for example be a plain JavaScript object containing node/row/document data from a database call. This source type data is in turn passed down to all the object type's own field resolvers.
+
+Here is the above diagram updated to include source types now.
+
+![](/assets/source-type-diagram-2.png)
+
+Here is a step-by-step breakdown of what is going on (follow the diagram annotation numbers):
```
jasonkuhrt, line 29: Diagram is currently missing annotation numbers right?

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+
+1. Client sends a query
+2. The field resolver for `Query.user` runs. Remember `Query` fields (along with `Subscription`, `Mutation`) are _entrypoints_.
```
jasonkuhrt, line 32:
```
_entrypoints_, also known as _root fields_ because they are on [_root types_](https://graphql.org/learn/execution/#root-fields-resolvers).
```

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+// ...
+}
+
+export interface Comment {
+// ^-------------^---------- Export your source type (required)
+//               |---------- Name of your source type which will be extracted by Nexus
+// ...
+}
+
+// ./schema.ts
```
jasonkuhrt, line 196:
```suggestion
// ./schema.ts

```

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+import { makeSchema, objectType, queryType } from '@nexus/schema'
+import * as path from 'path'
+
+const User = objectType({
+  name: 'User',
+  //      ^---------------------- Name of your GraphQL Object type which will be matched against the name of your source types
```
jasonkuhrt, line 202:
```suggestion
  //      ^---------------------- Name of your GraphQL Object type which will be matched
  //                              against the name of your source types
```

To fit website

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+    })
+  }
+})
+
+const Query = queryType({
+  definition(t) {
+    t.list.field('users', {
+      type: User,
+      resolve(_parent, args, ctx) {
+        //                       ^------- return as: User[]
```
jasonkuhrt, line 218:
```suggestion
        // ...                   ^------- return as: User[]
```

Optimize vertical lines?

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+        // ...
```
jasonkuhrt, line 219:
```suggestion
```

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+})
+
+const schema = makeSchema({
+  types: [User, Query],
+  typegenAutoConfig: {
+    sources: [{
+      source: path.join(__dirname, 'db.ts'),
+      //       ^----------------------------- Path to the file containing your source types
+      alias: 'db'
+      //      ^------------------------------ Arbitrary unique name used as alias for Nexus to import your source types. eg: import * as <alias> from <source>
```
jasonkuhrt, line 232: Let's try to keep within bounds of codeblock:

image

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+      //                                      Make sure to use unique names if you have multiple sources
+    }]
+  }
+})
+```
+
+In the example above, since the source type `interface User` matches the name of the `User` GraphQL Object type, Nexus will use the `interface User` as source type for the `User` GraphQL Object type.
+
+#### Different names following a convention
```
jasonkuhrt, line 241: I wonder if we can frame these as a set of strategies. That would align with how we talk about abstract types.

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+  typegenAutoConfig: {
+    sources: [{
+      source: path.join(__dirname, 'db.ts'),
+      //       ^----------------------------- Path to the file containing your source types
+      alias: 'db',
+      //      ^------------------------------ Name used as alias for Nexus to import your source types. eg: import * as <name> from <path>
+      typeMatch(type, defaultRegex) {
+      //         ^--------^---------------- GraphQL Named type to match against your source types
+      //                  |---------------- Default regex used to match your GraphQL types against your source types
+        return new RegExp(`(?:interface|type|class|enum)\\s+(${type.name}Model)\\W`, 'g')
```
jasonkuhrt, line 315: It took me a while to realize this is regex run literally against source code syntax. I think we need to make that a bit clearer here? 

Probably in a few places.

For example above:

>  to teach Nexus about your convention using regexes

Maybe:


> to teach Nexus about your convention. The way it works is that you give regular expressions to Nexus which it will run against the literal source code within modules (also specified by you) to find which types should be used for source types. Your regular expression must return a regex capturing group containing the TS type name of the source type. Here is a complete and annotated example to illustrate the system:


ish?

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+      //            ^---------------------- New regex which will match `UserModel` with `User`, etc..
+      },
+    }]
+  }
+})
+```
+
+You can also use multiple regexes to map your source types by returning an array of regexes:
```
jasonkuhrt, line 323: If multiple match, which wins?

docs/content/014-guides/022-source-types.mdx
```diff
@@ -0,0 +1,449 @@
...
+          //    ^---------------- Pass in the default Regex to keep the default behavior
+          new RegExp(`(?:interface|type|class|enum)\\s+(${type.name}Model)\\W`, 'g'),
+          // new RegExp(...), -> Add more regexes if you have even more convention
+        ]
+      },
+    }]
+  }
+})
+```
+
```
jasonkuhrt, line 343: What happens when the regex does not match (or none of the multiple given do)? Error? Any?

jasonkuhrt (approved):
Wow so simple and clear, amazing work @Weakky! Approved but consider some of the comment before merging probably :D

docs/content/014-guides/022-source-types.mdx
Weakky: image

Thanks for suggestion 🙏

docs/content/014-guides/022-source-types.mdx
Weakky: First one! Updated 👍

PR #712, opened 2020-12-07
chore(docs): update "Why Nexus"

chenkie:


docs/content/010-getting-started/04-why-nexus.mdx
```diff
@@ -2,18 +2,119 @@
...
+      body: 'How I wrote my first GraphQL server',
+    },
+  ],
+}
+```
 
-By combining automatic type generation with some of the more powerful features of TypeScript - type merging, conditional types, and type inference, we can know exactly which type names we are referring to and able to use throughout our code. We can know both the parameters and the return type of resolvers without providing any type annotation. It takes a little getting used to, but it ends up leading to a great feedback loop of the types annotating themselves.
+The schema type definitons, along with the resolvers, would then be passed to something like `ApolloServer` to create and serve the API.
+
+This way of building a GraphQL server is often referred to as the "schema-first" approach. We start by defining the shape of our API and then write code to tell it out to return data.
```
jasonkuhrt, line 39:
```suggestion
This way of building a GraphQL server is often referred to as the "schema-first" approach. We start by defining the shape of our API and then write code to tell it how to return data.
```
?

docs/content/040-adoption-guides/040-nexus-with-nextjs.md
```diff
@@ -0,0 +1,328 @@
+---
+title: Nexus With Next.js
```
jasonkuhrt, line 2: This shouldn't be here :)

jasonkuhrt (commented):
Nice! 

If I recall you still want to add a section about graphql relay connections and the plugins (e.g. standardize on descriptions).

PR #713, opened 2020-12-08
feat: add support for interfaces implementing interfaces to sdlConverter

danielrearden:
#496 added support for interfaces implementing interfaces, but these changes were not reflected in the SDL converter. This PR adds support for that there as well :)


tgriesser:
Nice, thanks @danielrearden!

Issue #647, opened 2020-11-20
Refactor configuration


jasonkuhrt:
Current typegen related options:

```
outputs.typgen
shouldGenerateArtifacts
typegenAutoConfig
typegenConfig
prettierConfig
formatTypegen
```

Suggestion:

```ts
sdl: {
  enabled: boolean // shouldGenerateArtifacts
  path: string
}
typegen: {
  enabled: boolean // shouldGenerateArtifacts
  path: string // outputs.typegen
  format: prettierConfig | formatTypegen
  createInfo: ... // typegenConfig
  headers: ... // typegenAutoConfig.headers
}
typings: {
  context: ...
  backing: { // ...typegenAutoConfig - contextType
    sources: ...
    map:... // typegenAutoConfig.backingTypeMap
    skip: ...
    debug: boolean
  }
}
```

jasonkuhrt:
WDTY @tgriesser @Weakky

jasonkuhrt:
I think `typings.backing` could be renamed `typings.models`.

Weakky:
I like it a lot overall. The part I'm not sure about is `sdl`. SDL is not a newcomer-friendly name. The problem is that I don't see `schema` being a better name, as it feels like it's configuring some stuff related to the whole library. That's probably the reason you've chosen `sdl` in the first place.

As we discussed, I also prefer `models` over `backing` as it's a much more familiar naming that we use other domains for similar concepts (such as orms). I never heard of the term "backing type" before working on @nexus/schema.

jasonkuhrt:
> That's probably the reason you've chosen sdl in the first place.

Yep

> The part I'm not sure about is sdl. SDL is not a newcomer-friendly name. 

`GraphQLSchemaDefinitionLanguageFile` is a verbose version, better?

This option is enabled by default now so it might not matter as much as before to be a newcomer friendly term.

Also there is always jsdoc fwiw.

tgriesser:
Going to need to think on this one for a bit

jasonkuhrt:
IMO I don't think this can be done after 1.0

jasonkuhrt:
Updated to use "source" terminology

jasonkuhrt:
Via @Weakky related to https://github.com/graphql-nexus/schema/issues/583#issuecomment-740046565 wanting to sync local type def `sourceTyping` with naming we have globally.

Updated to remove `typings` property namespace.

tgriesser:
[...]

PR #716, opened 2020-12-08
feat: make contextType independant of source types

Weakky:
BREAKING CHANGES:
- `contextType` is no longer nested in `typegenAutoConfig`
- `contextType` can no longer references `sources` registered in `typegenAutoConfig`
- `contextType` API was refactored to use  `module` instead of `path` and `export` instead of `name`
- `rootTyping` API was refactored to use  `module` instead of `path` and `export` instead of `name`


jasonkuhrt (approved):
Relates to #647 

LGTM

Let's see what @tgriesser thinks

tests/backingTypes.spec.ts
```diff
@@ -59,7 +59,10 @@ describe('backingTypes', () => {
             source: path.join(__dirname, '_types.ts'),
           },
         ],
-        contextType: 't.TestContext',
+      },
+      contextType: {
+        path: path.join(__dirname, '_types.ts'),
+        name: 'TestContext',
```
Weakky, line 65: Wondering if the type of `contextType` shouldn't be instead:

```ts
{
  module: string // instead of path
  name: string // name could be `export` here
}
```

Having `module` would be more consistent with the next PR which will rename `typegenAutoConfig.sources` `sourceTypes.modules`

jasonkuhrt: Agreed, and JS doc will make it clear that it is a path.

I personally find `name` not good either. Would suggest `import` instead.

Verbose would be `modulePath` and `importName`. But I think `module`/`import` are enough.

jasonkuhrt: It could be argued that we should use the import syntax as inspiration

```ts
import { foo } from 'bar'
```

```ts
{
  import: 'foo',
  from: 'bar'
}
```

:)

jasonkuhrt: Actually I think this might be best:

```ts
{
  module: 'bar'
  export: 'foo',
}
```

Since its specifying how to import, which is different than being an import.

```ts
import { foo } from 'bar'
//                   ^the module to import from
//       ^the export to import
```

:)

Weakky: done 👍

PR #718, opened 2020-12-09
chore(docs): use plain JS objects for storing the context

shreyas44:
- Use a JS object to store the GraphQL resolver context instead of a function that returns the object.

### DIFF
--- a/docs/content/010-getting-started/03-tutorial/04-chapter-3-adding-mutations-to-your-api.mdx
+++ b/docs/content/010-getting-started/03-tutorial/04-chapter-3-adding-mutations-to-your-api.mdx
@@ -106,6 +106,10 @@ export const server = new ApolloServer({
 
 </TabbedContent>
 
+The context passed to Apollo Server can either be a function or a plain JavaScript object. For this tutorial, we've used an Object for simplicity. We would need to use a function when we need to perform some calculation or get some data from an external source based on the request and then pass that derived data on to the context for all our resolvers to use. 
+
+For example, if a user is logged in to your application, it would be useful to have the information regarding the user available to all the resolvers and this information would have to be retrieved from your database or an external service. The code responsible for that database or service call would be placed in the function which is passed to Apollo Server as the context.
+
 Finally, let's configure Nexus to know the type of our GraphQL context by adjusting the configuration of the `makeSchema` in our `api/app.ts`.
 
 <TabbedContent tabs={['Diff', 'Code']}>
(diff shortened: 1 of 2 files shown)


docs/content/010-getting-started/03-tutorial/04-chapter-3-adding-mutations-to-your-api.mdx
```diff
@@ -63,10 +63,8 @@ export interface Context {
...
 }
 
-export function createContext(): Context {
-  return {
-    db
-  }
+export const context = {
+  db
 }
 ```
```
shreyas44, line 70: Should we also add a note that we can pass a function for the context, and why we would have to do that in some cases?

jasonkuhrt: Yeah that would be nice

jasonkuhrt (approved):
Will just let you add that extra info then will merge

docs/content/010-getting-started/03-tutorial/04-chapter-3-adding-mutations-to-your-api.mdx
```diff
@@ -95,19 +93,23 @@ export const server = new ApolloServer({
...
-  context: createContext
+  context
 })
 ```
 
 </tab>
 
 </TabbedContent>
 
+The context passed to Apollo Server can either be a function or a plain JavaScript object. For this tutorial, we've used an Object for simplicity. We would need to use a function when we need to perform some calculation or get some data from an external source based on the request and then pass that derived data on to the context for all our resolvers to use.
```
shreyas44, line 109: Done! Let me know if the placement and content is alright.

docs/content/010-getting-started/03-tutorial/04-chapter-3-adding-mutations-to-your-api.mdx
```diff
@@ -95,19 +93,23 @@ export const server = new ApolloServer({
...
 ```
 
 </tab>
 
 </TabbedContent>
 
+The context passed to Apollo Server can either be a function or a plain JavaScript object. For this tutorial, we've used an Object for simplicity. We would need to use a function when we need to perform some calculation or get some data from an external source based on the request and then pass that derived data on to the context for all our resolvers to use. 
+
+For example, if a user is logged in to your application, it would be useful to have the information regarding the user available to all the resolvers and this information would have to be retrieved from your database or an external service. The code responsible for that database or service call would be placed in the function which is passed to Apollo Server as the context.
+
```
shreyas44, line 112: Done! Let me know if the placement and content is alright.

jasonkuhrt:
Thanks @shreyas44!

Sytten:
Having a function to create the context is a very common pattern, I don't think I ever saw a production server with a static context.

shreyas44:
@Sytten If you have a dynamic context which is derived or calculated from say the content of the http request, a function is definitely the right way to go. But, if it's static (which in the case of the tutorial it is) I don't think it makes a difference if you use an object or a function. I thought passing a plain object might make more sense in this case as there's no dynamic context at all.

jasonkuhrt:
The tutorial is about starting from nothing and building up out of necessity to learn. This change is more inspirit with that. The way I would address your point would be to have a the chapter about auth show why a function for context is needed.

PR #719, opened 2020-12-09
fix(toc): marker outside of li, fix positioning

janpio:


Weakky (approved):
> So smart he doesn't even need to fire up his IDE to fix CSS

That'll be my linkedin recommendation when I'll leave Prisma ❤️

PR #743, opened 2020-12-18
chore(docs): fix examples for abstract types to work with Typescript

huv1k:
Fix docs for abstract types to work with Typescript. With unions, it's not possible to use properties that are not intersections between these types. We can use `in` operator to have it working with Typescript.


jasonkuhrt (approved):
ty!

PR #751, opened 2020-12-21
chore(docs): Fix missing export

kunalshah:
Address the error "Property 'export' is missing in type '{ module: string; alias: string; }' but required in type 'TypingImport'.ts(2741)"


jasonkuhrt (approved):
ty!

PR #762, opened 2020-12-31
Add missing import of db client to code diff in testing tutorial

TLadd:
Add missing import of db client to code diff in testing tutorial

The `Code` tab has this import but the diff tab is missing it and doesn't make sense without it.


TLadd:
Duplicate of https://github.com/graphql-nexus/nexus/pull/755. This caused some confusion for me; would be nice to have it merged.

PR #766, opened 2021-01-01
fix(docs): update 07-chapter-6-testing-with-prisma.mdx

canrozanes:
current implementation `prisma migrate up --create-db --experimental` throws the following error:

```
Command failed: /Users/canrozanes/Projects/ikiler-file-manager/node_modules/.bin/prisma migrate up --create-db --experimental
    Error: The current command "up" doesn't exist in the new version of Prisma Migrate.
    Read more about how to upgrade: https://pris.ly/d/migrate-upgrade
```

Changing the command to `prisma migrate dev --preview-feature` fixes the problem and runs the tests and passes them.


jasonkuhrt (approved):
ty!

PR #774, opened 2021-01-06
feat: allow implementing interfaces in extendType

### LINKED ISSUE #773: cannot implement interfaces in extendType definition
reproduction code: https://codesandbox.io/s/nexus-impelemens-in-extendtype-d8b0e?file=/src/schema/index.ts

I got a type error when using `implements` in `extendType`'s 'definition' block.

```ts
const addIDToUser = extendType({
  type: 'User',
  definition(t) {
    t.implements('Node')
    // ^^^^^^^^^
    // Property 'implements' does not exist on type 'OutputDefinitionBlock<"User">'.ts(2339)
  }
})
```

but generated schema has `type User implements Node` correctly.

```graphql
type User implements Node {
  id: ID!
  name: String!
}
```

izumin5210:
to fix #773 

Allow implementing interfaces in `extendType` definition

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -906,12 +906,6 @@ export class SchemaBuilder {
       warn: consoleWarn,
     })
     config.definition(definitionBlock)
-    const toExtend = this.typeExtendMap[config.name]
-    if (toExtend) {
-      toExtend.forEach((e) => {
-        e.definition(definitionBlock)
-      })
-    }
 
     if (config.sourceType) {
       this.rootTypings[config.name] = config.sourceType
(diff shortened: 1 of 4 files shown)


src/builder.ts
```diff
@@ -906,12 +906,6 @@ export class SchemaBuilder {
       warn: consoleWarn,
     })
     config.definition(definitionBlock)
-    const toExtend = this.typeExtendMap[config.name]
-    if (toExtend) {
-      toExtend.forEach((e) => {
-        e.definition(definitionBlock)
-      })
-    }
```
izumin5210, line 914: currently, we cannot extend interface types, so it's safe to remove these codes
https://codesandbox.io/s/extendtype-and-interface-ysf1v?file=/src/schema/index.ts

PR #777, opened 2021-01-06
chore(docs): remove outdated prisma plugin config docs

ahmedosama5200:
@jasonkuhrt Remove the [plugin settings section](https://nexusjs.org/docs/plugins/prisma/overview#plugin-settings) entirely. It contains more deprecated code like using `client` property which is now replaced by `prismaClient` of type `PrismaClientFetcher`

In addition, It's already covered in [configuration section](https://nexusjs.org/docs/plugins/prisma/overview#configuration) above.


jasonkuhrt (approved):
ty!

PR #781, opened 2021-01-12
docs: fix code example

denisborovikov:
1. Code and diff tabs code don't match
2. This is the first appearance of `stringArg`, it's never used before.
3. Display SDL code in a tab (same manner as the existing code below)


jasonkuhrt (approved):
ty!

PR #795, opened 2021-01-20
fix: Allow specifying overrides for builtin scalars

tgriesser:
If we provide a custom type for a builtin scalar type in the `sourceTypes.mapping`, it should override the default mapping. 

This is specifically useful if you wish to create a pseudo [nominal type](https://basarat.gitbook.io/typescript/main-1/nominaltyping) for `ID` to distinguish from a normal string.


jasonkuhrt (approved):
LGTM

PR #802, opened 2021-01-26
docs: Fix variable naming for express import

Zeko369:


tgriesser:
Yeah, this seems to be more how I'd expect it.

PR #808, opened 2021-01-29
chore(docs): Correct an object's value in the tutorial

yashguptaz:
In the ContextType Object, the value of `export` should be `Context` instead of `ContextModule` (see Line 133)
On line 62, the name of the exported interface is `Context`.

I have noticed that autocompletion of Prisma doesn't work because of the misspelling :)

When I checked the repository of the tutorial the value of `export` was correct.
https://github.com/graphql-nexus/tutorial/blob/e7a1988eded47a818398d14d823e45423ddeb3dd/api/schema.ts#L12


jasonkuhrt (approved):
ty!

PR #809, opened 2021-01-29
chore(docs): Fix typo in About section

johnnyoshika:
From: It gives you an API you to project fields from models defined in your Prisma schema
To: It gives you an API ~~you~~ to project fields from models defined in your Prisma schema


jasonkuhrt (approved):
Ty!

Issue #549, opened 2020-10-13
Improve fieldAuthorizePlugin documentation (formatError option)


MikaStark:
It's seems you can only get "**Not authorized**" error even when you throw or return a custom error. As the documentation for fieldAuthorizePlugin is really thin, is there a way to return/throw custom error ?

If you use _graphql-shield_, returning a string or a custom error do the job.

Weakky:
Hey @MikaStark, you can use the `fieldAuthorizePlugin({ formatError })` option to configure the error returned.

The default `formatError` implementation is the following:

```ts
export const defaultFormatError = ({ error }: FieldAuthorizePluginErrorConfig): Error => {
  const err: Error & { originalError?: Error } = new Error('Not authorized')
  err.originalError = error
  return err
}
```

You can customize it to whatever you want instead

```ts
fieldAuthorizePlugin({
  formatError(error) {
    // customize it as you wish or just return the original error
  }
})
```

I'll keep the issue opened and rename the title to make it about docs

MikaStark:
Hello @Weakky, thank you for your explanation, it works perfectly this way :). But why don't just expose the error by default whenever it exists ?

```ts
export const defaultFormatError = ({ error }: FieldAuthorizePluginErrorConfig): Error => {
  return error ?? new Error('Not authorized')
}
```

Weakky:
We'd have to ask @tgriesser for that. Is that maybe to prevent leak security info about why it's not authorized?

MikaStark:
In this case, adding (like graphql-shield) a optional option that allow/disallow external errors could be nice (like ones thrown by Prisma, DataSources, etc.). By default this option will be true in dev mode and false in production.

LRipi:
Hello @Weakky ,
Maybe there's something I didn't understand, but I still have a problem with the plugin.
To make it simple, I format the error perfectly in the "formatError" function, it does return the error containing what was thrown in my Query fields.
But, for a reason unknown to me, when Apollo-server will get this error, it will always be "Not authorized", whatever the thrown error.

Is there anyone who can help me?

Here is my code:
```
  const schema = makeSchema({
    outputs: {
      schema: path.resolve('src/generated/schema.graphql'),
      typegen: path.resolve('src/generated/typings.ts')
    },
    plugins: [
      queryComplexityPlugin(),
      connectionPlugin(),
      fieldAuthorizePlugin({
        formatError: ({ error }) => {
          console.log(error);
[...]

PR #821, opened 2021-02-08
docs: add more information to field authorize plugin

reconbot:
Added some getting started information and a description of the field this plugin adds.

works #549


tgriesser:
Thanks!

Issue #844, opened 2021-02-28
Limited configurability of connectionPlugin connection and edge types


vecerek:
In `@nexus/schema@0.13.1` it was possible to express a schema like:
```gql
type ResourceConnection {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Edge-Types
  """
  edges: [ResourceEdge]

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
  """
  pageInfo: PageInfo!
}

type ResourceEdge {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Cursor
  """
  cursor: String!

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Node
  """
  node: Resource!
}
```

After upgrading to `nexus@1.0.0` I found that this schema is no longer achievable.

With `nonNullDefaults: { output: false }` I get:

```gql
type ResourceConnection {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Edge-Types
  """
  edges: [ResourceEdge]

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
  """
  pageInfo: PageInfo!
}

type ResourceEdge {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Cursor
  """
  cursor: String!

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Node
  """
  node: Resource
}
```

with `nonNullDefaults: { output: true }`:

```gql
type ResourceConnection {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Edge-Types
  """
  edges: [ResourceEdge!]!

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
  """
  pageInfo: PageInfo!
}

type ResourceEdge {
  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Cursor
  """
  cursor: String!

  """
  https://facebook.github.io/relay/graphql/connections.htm#sec-Node
  """
  node: Resource!
}
```

tgriesser:
If I understand correctly, what you're after is for is a way to specify `Resource` as non-null, is that correct?

If so, you should be able to achieve this by doing:

```ts
import { nonNull } from 'nexus'

// ...

t.connectionField('resource', {
  type: nonNull(Resource)
})
```

vecerek:
[...]

PR #845, opened 2021-02-28
fix: make connectionPlugin compliant with Relay Connections spec

vecerek:
Fixes: #844 

The Relay Connections specification says:

> If ExampleConnection existed in the type system, it would be a connection, since its name ends in “Connection”. If this connection’s edge type was named ExampleEdge, then a server that correctly implements the above requirement would accept the following introspection query, and return the provided response:

```gql
{
  __type(name: "ExampleConnection") {
    fields {
      name
      type {
        name
        kind
        ofType {
          name
          kind
        }
      }
    }
  }
}
```

> returns

```json
{
  "data": {
    "__type": {
      "fields": [
        // May contain other items
        {
          "name": "pageInfo",
          "type": {
            "name": null,
            "kind": "NON_NULL",
            "ofType": {
              "name": "PageInfo",
              "kind": "OBJECT"
            }
          }
        },
        {
          "name": "edges",
          "type": {
            "name": null,
            "kind": "LIST",
            "ofType": {
              "name": "ExampleEdge",
              "kind": "OBJECT"
            }
          }
        }
      ]
    }
  }
}
```

https://relay.dev/graphql/connections.htm#sec-Connection-Types.Introspection

Furthermore, it also says:

> An “Edge Type” must contain a field called node. This field must return either a Scalar, Enum, Object, Interface, Union, or a Non‐Null wrapper around one of those types. Notably, this field cannot return a list.

https://relay.dev/graphql/connections.htm#sec-Node

The nullability of these fields can currently be controlled in a limited way through the `nonNullDefaults` configuration option.  I believe the specification can not be satisfied with the current limitations around configurability. This PR makes sure the connection plugin is compliant with the specification. If consumers need to use non-compliant connection types, they should implement their own plugin/types until the limitations around configurability are addressed.

### DIFF
--- a/tests/plugins/__snapshots__/connectionPlugin.spec.ts.snap
+++ b/tests/plugins/__snapshots__/connectionPlugin.spec.ts.snap
@@ -200,7 +200,7 @@ exports[`basic behavior should adhere to the Relay spec 2`] = `
   cursor: String!
 
   \\"\\"\\"https://facebook.github.io/relay/graphql/connections.htm#sec-Node\\"\\"\\"
-  node: User
+  node: User!
 }"
 `;
 
@@ -303,7 +303,7 @@ type QueryFieldLevel_Connection {
   \\"\\"\\"
   https://facebook.github.io/relay/graphql/connections.htm#sec-Edge-Types
   \\"\\"\\"
-  edges: [QueryFieldLevel_Edge!]!
+  edges: [QueryFieldLevel_Edge]
 
   \\"\\"\\"
   https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
@@ -324,7 +324,7 @@ type QueryFieldLevel2_Connection {
   \\"\\"\\"
   https://facebook.github.io/relay/graphql/connections.htm#sec-Edge-Types
   \\"\\"\\"
-  edges: [QueryFieldLevel2_Edge!]!
+  edges: [QueryFieldLevel2_Edge]
 
   \\"\\"\\"
   https://facebook.github.io/relay/graphql/connections.htm#sec-undefined.PageInfo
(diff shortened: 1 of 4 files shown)


tests/plugins/__snapshots__/connectionPlugin.spec.ts.snap
```diff
@@ -200,7 +200,7 @@ exports[`basic behavior should adhere to the Relay spec 2`] = `
   cursor: String!
 
   \\"\\"\\"https://facebook.github.io/relay/graphql/connections.htm#sec-Node\\"\\"\\"
-  node: User
```
vecerek, line 203: I'm wondering why these snapshots were incorrect in the first place. This definitely does not adhere to the Relay Connection specification. There's also a very good reason behind the design choices in the spec.

```gql
edges: [User!]!
...
cursor: String!
node User!
```

The above is bad because an issue with a node, e.g. a required field resolves to `null`, results in a null value propagating up to the first non-nullable field. And because edges can't have null elements and edges themselves can't be null, one broken node would break the entire connection.

```gql
edges: [User!]!
...
cursor: String!
node User
```

In this case, node can be nullable but then a non-nullable cursor doesn't make any sense. `cursors` are usually derived from the node and must be unique in the connection. How does one derive the cursor for a `null` and what if there are 2 `null` nodes in the result set?

The only configuration that makes sense is the specified:
```gql
edges: [User]
...
cursor: String!
node User!
```
possibly with edges being `[User]!` in my opinion.

tgriesser:
Commented separately on #844 with a solution for your use case. 

The current behavior is intentional, and I do not plan on hardcoding the nullability in the connection plugin, as every project has its own tradeoffs with nullability guarantees. 

Out of the box, without configuring `nonNullDefaults`, the connection plugin should be spec compliant, and there is an option to configure `nonNullDefaults` specifically for connections in the plugin configuration.

vecerek:
> Out of the box, without configuring `nonNullDefaults`, the connection plugin should be spec compliant, and there is an option to configure `nonNullDefaults` specifically for connections in the plugin configuration.

@tgriesser thanks for the quick answer. Is it intended that the `connectionPlugin` respects the `nonNullDefaults` configuration as specified in `makeSchema` options?

```ts
export const schema = makeSchema({
  nonNullDefaults: { input: true, output: true },
  plugins: [connectionPlugin()], // is the above configuration inherited by this plugin?
  ...
});
```

PR #847, opened 2021-03-02
chore(docs): update 010-object-type.mdx

AntonNiklasson:
Looks like there should only be a single argument passed to `objectType`.


jasonkuhrt (approved):
ty!

PR #859, opened 2021-03-14
chore(docs): Change --transpileOnly to --transpile-only from star-wars example

nilportugues:
ts-node-dev will not start the process if --transpileOnly is passed as it is deprecated.


jasonkuhrt:
So how do you skip type checking?

nilportugues:
ts-node-dev won't run that flag as camelCase but will work fine if kebab-case.

PR #880, opened 2021-04-08
chore(docs): Nexus nullability defaults to that everything is null by default

ahmedosama5200:
Nexus nullability defaults to that everything is null by default. It's kind of the opposite is done here


jasonkuhrt (approved):
ty!

Issue #883, opened 2021-04-09
generated artifacts work differently than described in the docs


astroash:
The documentation mentions that generated artifacts are created when not in production:

> By default, this only occurs when process.env.NODE_ENV !== "production".

However looking at this libraries code, artifacts are only generated if there is no env or it is `developement`. This does not allow consumers of the library to define their own environment names.

https://github.com/graphql-nexus/nexus/blob/7b84cd63db525fd7a832da1dee1900aabefbac59/src/typegenUtils.ts#L10

astroash:
fix by https://github.com/graphql-nexus/nexus/pull/884

PR #884, opened 2021-04-09
fix: sync should gen artifact logic with what docs say

astroash:
…ather than only when in development

 #883


jasonkuhrt:
@tgriesser I'm fine with this change, any objections? It is technically a breaking change. Maybe we should just change the docs.

astroash:
@jasonkuhrt just for reference, this change is in line with how the documentation describes generated artifacts

https://nexusjs.org/docs/guides/generated-artifacts#introduction

jasonkuhrt:
@astroash true, but might be different than observed behaviour so far.

tgriesser:
Yeah this change seems fine.

PR #904, opened 2021-05-12
chore(docs): fix proper variable name

thilllon:
Change the variable name appropriately in the example code
https://nexusjs.org/docs/api/query-field


jasonkuhrt:
ty!

PR #906, opened 2021-05-15
Fix maybe deep promise type

### LINKED ISSUE #853: Allow promises to be returned in a field in the resolver of connection
The below example would return a type error because the return type of the edges field is expected to be `{cursor: string, node: Bar}` and `Promise<{cursor: string, node: Bar}>` is not allowed. However, a query that requests for those edges will resolve without any error.

**Why return a Promise in the getter of edges in the first place?**

Say, we need to get a list of items from a database and parse that through a utility function that creates a cursor for that record. If we directly `await` the query to get the list of items, then we lose the ability to batch the requests using something like dataloader, or Prismas Fluent API (which uses the dataloader pattern). Instead, we can return a promise for the edges field that waits for the response from the database, parse that through our utility function to get the cursor, and resolve the promise.

An alternative is to just use the `nodes()` function, but then we have to query for all the fields in the database (as the `pageInfo` is generated from returned value of `nodes()`, and this is really not viable if you have a huge number of records.

```ts
import { ApolloServer } from "apollo-server"
import { connectionPlugin, makeSchema, objectType, queryField } from "nexus"

const bars = [
  {
    cursor: "cursor1",
    node: {
      id: "1",
      barField1: "abcd",
      barField2: "abcdef",
    }
  },
  {
    cursor: "cursor2",
    node: {
      id: "2",
      barField1: "asdf",
      barField2: "asdfasdf",
    }
  }
]

const Foo = objectType({
  name: "Foo",
  definition(t) {
    t.nonNull.id("id")
    t.nonNull.string("fooField1")
    t.nonNull.string("fooField2")
   
   // --- Error in this part ---
    t.connectionField("bars", {
      type: "Bar",
      resolve() {
        return {
          get edges() {
            return new Promise(resolve => {
              resolve(bars)
            })
          }
        }
      }
    })
  }
})

const Bar = objectType({
  name: "Bar",
  definition(t) {
    t.nonNull.id("id")
    t.nonNull.string("barField1")
    t.nonNull.string("barField2")
  }
})

const fooQuery = queryField("foo", {
  type: "Foo",
  resolve() {
    return {
      id: "12345",
      fooField1: "qwrew",
      fooField2: "14123"
    }
  }
})

const schema = makeSchema({
  types: [Foo, Bar, fooQuery],
  plugins: [connectionPlugin()]
})
```

I'm sorry for the bad example, let me know if you'd like a better one 🙂

shreyas44:
Removing the conditional type for null in `MaybePromiseDeep` seems to fix the below linked issue without removing any type safety.
I've tested it with the getters for `edges` and `pageInfo` with and without promises.

However, I'm not entirely sure if this breaks something else as it seems to be a utility used in mutliple places across the library.

Closes #853﻿

### DIFF
--- a/src/typegenTypeHelpers.ts
+++ b/src/typegenTypeHelpers.ts
@@ -1,4 +1,5 @@
 import type { GraphQLAbstractType, GraphQLResolveInfo } from 'graphql'
+
 import type { NexusInterfaceTypeDef } from './definitions/interfaceType'
 import type { NexusObjectTypeDef } from './definitions/objectType'
 
@@ -35,8 +36,6 @@ export type MaybePromise<T> = PromiseLike<T> | T
  * use this to help signify that.
  */
 export type MaybePromiseDeep<T> = Date extends T
-  ? MaybePromise<T>
-  : null extends T
   ? MaybePromise<T>
   : boolean extends T
   ? MaybePromise<T>


jasonkuhrt:
Is the change in this PR causing https://github.com/graphql-nexus/nexus/pull/906/checks?check_run_id=2589178499#step:6:12? Would seem so but not sure.

shreyas44:
Yep, you're right, the change in the PR did cause the issue. It seems to stem from the fact that [TypeScript distributes unions when evaluating conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types). 

I fixed that and also removed a redundant `MaybePromise` type from the return type of `resolve` as `MaybePromiseDeep` returns a `MaybePromise` itself.

jasonkuhrt:
Tests still seem to be failing though?

PR #910, opened 2021-05-20
chore(tests): update snapshots that were forgotten in #887

janpio:
#887 was merged with failing tests, now the main branch is also failing.


janpio:
I think we can ignore this false positive from `codecov` here 🙄

PR #911, opened 2021-05-20
chore: Remove unmaintained Changelog

### LINKED ISSUE #221: Changelog maintenance?
Hi! I notice you have a changelog in this project, but the most recent releases aren't included in it, and likewise, the most recent releases don't have any release notes, so it's hard to understand what's changing in Nexus over time, making updating more time-consuming.

Is there a plan around improving this? (For instance, you could use semantic-release or semantic-changelog to automatically generate it from git commits)

janpio:
The changelog has not been maintained since 0.18, but releases have been.

For people that still need it, it is present in the old tags.

closes #221


jasonkuhrt:
@tgriesser I think you actually brought this back at one point. Is this cool with you now?

tgriesser:
I forget where, but I think I'd asked if we could bring it back via dripip or some other automation, since I used to manually curate this when I used to do the releases by hand, but that went away when we switched to the new CI process.

I think it's important to have a single location to read through changes rather than needing to paginate through releases pages.

PR #916, opened 2021-05-24
feat: Accept `ReadonlyArray<T>` for enum members

### LINKED ISSUE #455: Can `enumType.members` allow `readonly string[]`?
I have the following data setup (not declaring as an enum because this is shared between TypeScript server code and non-TypeScript client code):

```js
import { enumType } from 'nexus';

export const TYPE_ENUM = Object.freeze({
  H: 'H',
  S: 'S',
});
export const TYPES = Object.freeze(Object.values(TYPE_ENUM));

export const MyType = enumType({
  name: 'MyType',
  members: TYPES,
});
```

TypeScript is rightly giving me an error here because I'm trying to put `TYPES` (a `readonly string[]`) into `members` which accepts a `string[]`. 

https://github.com/graphql-nexus/schema/blob/bf0df64ebd9810f9a458aceb8c5ff52518607825/src/definitions/enumType.ts#L41-L45

To overcome the type issue the best solution I have found so far is to unwrap and then re-create the array: `members: [...TYPES]`

Is there a better way to overcome the type issue? If not, could `enumType.members` be modified to allow `readonly` arrays, or does `members` need to be mutable?

tom-sherman:
Closes #455

I added in a new `isArray` util which just calls `Array.isArray` but correctly narrows the type of the input. This is to workaround the issue described [here](https://github.com/microsoft/TypeScript/issues/17002), the solution was taken from the since reverted https://github.com/microsoft/TypeScript/pull/39258

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -600,3 +600,11 @@ export function graphql15InterfaceType<T extends GraphQLInterfaceType>(
   }
   return type as T & { getInterfaces(): GraphQLInterfaceType[] }
 }
+
+// A function that is correctly typed to get arround a TypeScript issue.
+// See https://github.com/microsoft/TypeScript/issues/17002
+export function isArray<T>(
+  arg: T | {}
+): arg is T extends readonly any[] ? (unknown extends T ? never : readonly any[]) : any[] {
+  return Array.isArray(arg)
+}
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -601,8 +601,9 @@ export function graphql15InterfaceType<T extends GraphQLInterfaceType>(
   return type as T & { getInterfaces(): GraphQLInterfaceType[] }
 }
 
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -600,3 +600,11 @@ export function graphql15InterfaceType<T extends GraphQLInterfaceType>(
   }
   return type as T & { getInterfaces(): GraphQLInterfaceType[] }
 }
+
+// A function that is correctly typed to get arround a TypeScript issue.
+// See https://github.com/microsoft/TypeScript/issues/17002
```
jasonkuhrt, line 605:
```suggestion
/**
 * A specially typed version of `Array.isArray` to work around [this issue](https://github.com/microsoft/TypeScript/issues/17002).
 */
```

jasonkuhrt (approved):
ty!

PR #917, opened 2021-05-24
feat: Support isolated modules

### LINKED ISSUE #870: Support pulling source types from "export type" syntax
When `--isolate-modules` is enabled (in a Next.js app for example) it is required to use `export type` syntax.

image

Nexus doesn't support pulling from this pattern. Needs a regex update.

tom-sherman:
Closes #870


jasonkuhrt:
This does not solve #870. That issue is about a source types system limitation. Would you like to pitch this PR as an improvement anyways?

jasonkuhrt (changes requested):
This does not solve #870. That issue is about a source types system limitation. Would you like to pitch this PR as an improvement anyways?

tom-sherman:
@jasonkuhrt Oh wow, I completely misunderstood the issue! 😆 I'll close this then and we can discuss in the issue.

PR #927, opened 2021-06-09
refactor: move makeSchema into separate file from builder

tgriesser:
Minor, but thought this was simpler when coming back to the codebase after awhile


jasonkuhrt (approved):
💯

PR #939, opened 2021-06-25
chore(docs): improve scalar documentation

tobiaslins:
I was messing around for ours to get a custom scalar working. I simply forgot to list in in the `types` array while creating the schema. There was no mention about this in the docs.

Thanks!


jasonkuhrt (approved):
ty

Issue #148, opened 2019-06-01
Apollo Federation support


jferrettiboke:
[Apollo Federation](https://www.apollographql.com/docs/apollo-server/federation/introduction/) was open sourced a few days ago by Apollo team. I would like to know if there are plans to integrate Apollo Federation into Nexus in some ways. I just pretend to discuss deeper how this would look like.

I think the way that Nexus provides to extend types is similar to what Apollo Federation does. However, Apollo Federation allows us to create a gateway from different URLs (other GraphQL servers as services) and each URL is a service running independently. I would like to discuss this with some of you guys because doing things in a similar way is easier to maintain and it's better for teams at scale.

The good thing about this is that every team can use the programming language they want (I read somewhere that Apollo plans to expand to other programming languages), and using the gateway in Apollo Federation, all URLs can be merged at once easily and clients will see all graphs connected. Think this like a microservices pattern for GraphQL services.

Nexus is great, but if we try to implement everything within the same server, it will not provide freedom to teams since everything must be done in the same language because Nexus has no ways yet to connect easily to other GraphQL services. In an attempt to mitigate this, I created an [issue](https://github.com/prisma/nexus/issues/98) in `prisma/nexus` regarding plugins, but it is still in discussion.

Also, running a solo-server using serverless is not efficient and costs are higher. We use Zeit Now and we like to split everything as much as we can so that we are pretty sure that once a function is called, it only does one thing (but very well). This scales so well. With Apollo Federation, you can have as many servers as services you have, plus the server which acts as a gateway.

I think having something like this is a big plus. We can discuss this as much as you want, guys. I would be more than happy to contribute to the future and success of Yoga.

tgriesser:
Not yet, but I am interested in taking a look and seeing what the opportunity for integration might be here - been on vacation the last week or so, so I'm just getting caught up, haven't had a chance to take a look at the specifics of how Apollo Federation is spec'ed to work.

> I created an issue in prisma/nexus regarding plugins, but it is still in discussion.

Interested to see what you think of the changes merged in #143 and if it'll address anything you're looking to do re: plugins

Nayni:
After reading through the current Federation spec I think there is a great opportunity to build additional functionality on top of `nexus` to adhere to the Federation spec and allow building these types of schema's.

[...]

Issue #53, opened 2019-02-22
allow custom directives


ngmariusz:
if i understand correctly there is only @deprecated possible so far? it there way to add custom ones? 
tools out there usually depend on custom directives then i could integrate nexus into toolchain

tgriesser:
We can add the ability to define a directive for runtime use, but (as far as I can tell) adding them for use annotating the SDL would be more difficult, due to lack of support in graphql-js:
- https://github.com/graphql/graphql-js/issues/1262
- https://github.com/graphql/graphql-js/issues/1343

Can you talk more about the type of directives you are looking to add and how they would be used?

ngmariusz:
Sure, 

So Basically my question was motivated  with context the of trying to use Nexus  in already existing Project or pipeline (which is schema first) 
Let’s say i have a set of custom directives syntax that i use to annotate schema and then parse with my cli tools to generate resolvers, db and so on 
As Nexus actually generates schema  i’ll put it on the beginning of tool chain try to  achieve same result as i  would wright by hand in my .graphql files.
So finally i would just like to be able to achive in exp  this 

```
type User @entity {
  id: ID! @id
  email: String! @column @map(path: "login.email") @depreciated(reason: 'lorem')
}
```
with sth like 
```
const User = objectType({
 annotate:'@entity',
 name: 'User',
 definition(t) {
   t.string('email', { annotate:'@column @map(path: "login.email")', deprecation: 'lorem', nullable: true });
   t.id('id', { annotate:'@column' })
 }
});
```
without this feature switching to Nexus seems like I have to leave behind everything that already works

tgriesser:
I think this should be possible to add, going to dig into what it'll take to do this properly and keep you posted.

jregistr:
👋 Hellos! I'm also interested in this. Perhaps all we'd need for implementing this is the ability to specify directives for type and field creation? 
At a glance, looks like one can use [apollo's graph-tools](https://github.com/apollographql/graphql-tools) to make the schema while providing directives.

tgriesser:
👋 
> Perhaps all we'd need for implementing this is the ability to specify directives for type and field creation? At a glance, looks like one can use apollo's graph-tools to make the schema while providing directives.

Yes - you can add these in graphql-tools, but It's not quite as simple to do programmatically because they needed to be added to the AST for the schema, which is possible but takes a bit more work.

[...]

PR #952, opened 2021-07-21
feat: Add Schema Directives to SDL

### LINKED ISSUE #947: Support for operation directives
The discussion on issue #53 seems to be about schema directives, so I'm opening a new issue regarding query directives which is also not supported, 

It seems quite simple to add this feature by passing an extra field to the `GraphQLSchema` constructor here:
https://github.com/graphql-nexus/nexus/blob/8e0ce1b4a156cd50b061b47c8a1241729b4dd2a3/src/builder.ts#L1700-L1709

Since that is just one function down from `makeSchema` and uses the same type, all it requires is adding a new field on the `SchemaConfig` [here](https://github.com/graphql-nexus/nexus/blob/8e0ce1b4a156cd50b061b47c8a1241729b4dd2a3/src/builder.ts#L259-L277) and passing it along. I tested it locally, and it works perfectly. The schema gets generated correctly with the directive specification at the top, and the information is passed to the resolvers at `info.fieldNodes.directives`  as [graphql-js defines](https://github.com/graphql/graphql-js/blob/6a5f51f3bd07f3914c7f1ec01908e9e325b77d58/src/language/ast.ts#L304).

Here is how it ended up being with this change:
```typescript
/* file: directives.ts */
import {GraphQLDirective, GraphQLInt} from 'graphql';

export const cacheDirective = new GraphQLDirective({
    name: 'cached',
    args: {
        lifetime: {
            type: GraphQLInt,
        },
    },
    locations: ['FIELD'],
});

/* file: schema.ts */
import {makeSchema} from 'nexus';
import {cacheDirective} from './directives';

export const NexusGraphQLSchema = makeSchema({
    directives: [cacheDirective],
    ...
});

/* file: testResolver.ts */
import {queryField} from 'nexus';
import {cacheDirective} from '../directives';
import {getArgumentValues} from 'graphql/execution/values';

export const testField = queryField(def => def.string('testField', {
     resolve: (_, __, ___, info) => {
        const cacheConfiguration = getDirectiveValues(cacheDirective, info.fieldNodes[0], info.variableValues);
        return cacheConfiguration?.lifetime?.toFixed(0) ?? 'no-cache';
    },
});
```

It seems deceptively simple for no one to have done it by now, is there a problem with this that happens not to affect my application, or should I open a PR?

tgriesser:
Closes: #947, #53

Helps with: #148

WIP: Still needs docs / additional tests / cleanup of a bit of the builder layer / fixing types

Adds a `directives` property to all definition props which takes an array of directives. Adds them to an `astNode` which are then printed via a custom schema printer.

Definition:

```ts
const someDirective = directive({
  name: 'SomeDirective',
  locations: ['OBJECT'],
  args: {
    bool: nonNull(booleanArg())
  }
})
```

Use:

```ts
objectType({
  name: 'SomeType',
  directives: [someDirective({bool: true})]
})
```

or 

```ts
import { addDirective } from 'nexus'

objectType({
  name: 'SomeType',
  directives: [addDirective('SomeDirective', {bool: true})]
})
```

Prints:

```graphql
type SomeType @SomeDirective(bool: true) {
  # ... fields
}
```

See changes in the [`kitchenSink/__app.ts`](https://github.com/graphql-nexus/nexus/compare/tgriesser/feat/sdl-directives?expand=1#diff-4459f76e829acc4ad1fa32dc0267afba074f726199c749ef0be08ce3d2fa4fe7) for more example uses.


Sytten:
Dont close #148, still needs a plugin for it to make is nice for people to use (should be trivial once this is merged)

src/printSchemaWithDirectives.ts
```diff
@@ -0,0 +1,323 @@
...
+  isUnionType,
+  isEnumType,
+  isInputObjectType,
+} from 'graphql'
+
+import { astFromValue } from 'graphql'
+
+export function printSchemaWithDirectives(schema: GraphQLSchema): string {
+  return printFilteredSchemaWithDirectives(schema, (n) => !isSpecifiedDirective(n), isDefinedType)
+}
```
Sytten, line 36: I don't think we need a custom printer, all graphql versions that nexus support have a schema printer that does print the directives:
https://github.com/graphql/graphql-js/blob/v14.5.8/src/utilities/schemaPrinter.js
https://github.com/graphql/graphql-js/blob/v15.5.1/src/utilities/printSchema.js
And they should be exported by `graphql/utilities`.

src/makeSchema.ts
```diff
@@ -64,13 +65,61 @@ generateSchema.withArtifacts = async (
...
+    typeFilePath,
+    hasSDLDirectives
   )
   assertNoMissingTypes(schema, missingTypes)
   runAbstractTypeRuntimeChecks(schema, finalConfig.features)
   return { schema, schemaTypes, tsTypes }
 }
+
+/** Builds the schema, we may return more than just the schema from this one day. */
+export function makeSchemaInternal(config: SchemaConfig) {
```
Sytten, line 81: Do we want to export that anymore?

src/makeSchema.ts
```diff
@@ -64,13 +65,61 @@ generateSchema.withArtifacts = async (
...
+    query: Query,
+    mutation: Mutation,
+    subscription: Subscription,
+    types: objValues(typeMap),
+    extensions: {
+      ...config.extensions,
+      nexus: schemaExtension,
+    },
+    directives: [...specifiedDirectives, ...Object.values(customDirectives)],
+    ...schemaDirectives,
```
Sytten, line 119: I have a TS error in editor for that line `Property 'operationTypes' is missing in type 'ScalarTypeDefinitionNode' but required in type 'SchemaDefinitionNode'`

Sytten: I feel this is misleading for future changes, this returns an AST node. If some would need to override another property in the `astNode` they might override the `astNode` set by `schemaDirectives`.

src/builder.ts
```diff
@@ -869,6 +931,7 @@ export class SchemaBuilder {
         ...config.extensions,
         nexus: new NexusInputObjectTypeExtension(config),
       },
+      ...this.maybeAddDirectiveUses('INPUT_OBJECT', config.directives),
```
Sytten, line 934: I said it in the `makeSchema`, but I want to drive the point, since this adds a whole `astNode` I feel that someone that might want to modify the AST node for another purpose will get blocked by this code or might override it entirely. I think we should at the very least wrap that in an array and a merge function or even better if only the `directives` field of the node was returned by `maybeAddDirectiveUses` (free and member functions) and the `astNode` was generated by another method.

Sytten (changes requested):

src/definitions/args.ts
```diff
@@ -47,6 +48,13 @@ export type CommonArgConfig = {
    * graphql-js based tools which rely on looking for special data here.
    */
   extensions?: GraphQLScalarTypeConfig<any, any>['extensions']
+  /**
+   * A list of directives / directive uses (with args) for the arg definition
+   *
+   * @example
+   *   directives: [useDirective('ExampleDirective', { arg: true })]
+   */
+  directives?: Directives
```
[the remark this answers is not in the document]
tgriesser: Yeah I had initially tried this api as:
```ts
directives(t) {
  t.MyDirective(args)
  t.OtherDirective(args)
}
```

Where the members were dynamically based on directive declarations & where they can be positioned. It generally worked, but the types were pretty gnarly & seemed to be a lot of complexity for something that could just be a simple runtime error.

src/definitions/directive.ts
```diff
@@ -0,0 +1,211 @@
...
+  if (!directiveDef.locations.includes(kind)) {
+    throw new Error(`Directive ${directiveDef.name} cannot be applied to ${kind}`)
+  }
+}
+
+function directiveArgs(
+  directiveUse: NexusDirectiveUse,
+  directiveDef: GraphQLDirective
+): ReadonlyArray<ArgumentNode> {
+  return mapObj(directiveUse.args ?? {}, (val, key) => {
```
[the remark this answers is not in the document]
tgriesser: Hmm... Need to check on that

src/definitions/directive.ts
```diff
@@ -0,0 +1,211 @@
...
+   *         name: 'TestValueType',
+   *         members: ['String', 'Int', 'Float', 'JSON'],
+   *       }),
+   *       value: stringArg(),
+   *       listLength: intArg(),
+   *     },
+   *     locations: ['FIELD_DEFINITION'],
+   *   })
+   */
+  args?: ArgsRecord
```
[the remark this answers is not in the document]
tgriesser: No, good catch - need to do that

src/printSchemaWithDirectives.ts
tgriesser: I believe it only prints the directives declarations, not the directive uses on the SDL types/fields

src/makeSchema.ts
tgriesser: Yeah I think I'm using it in the tests, I don't believe it's publicly exported on the module

src/printSchemaWithDirectives.ts
Sytten: Good point, bit annoying that we have to copy paste/maintain all that code. We should reference the commit it was taken from and maybe the diff vs the original to reduce upgrade burden.

tgriesser: Yeah, I'm not too worried about it in the sense that since it's just printing code the paths are pretty simple and easy to validate correctness of (not a lot of business logic that you'd otherwise be worried about with copy/paste).

src/makeSchema.ts
tgriesser: Yeah agreed, meant to come back to this / out this as a TODO

src/builder.ts
tgriesser: > I feel that someone that might want to modify the AST node for another purpose will get blocked by this code or might override it entirely

Yeah, fair points - I had considered that and was thinking of just crossing that bridge if/when it came to it, since `astNode` isn't part of the public api for any of the nexus builder fn's, I wasn't sure if that would actually happen in practice.

Sytten: In theory no, we should never touch the AST node since it is supposed to the result of the parsing of the schema.
I still think it should be explicit in the code that we are converting the directives to an AST node. Something like:
```typescript
this.createAstNode([this.maybeAddDirectiveUses])
```
I suspect that a future release of graphql will add backend directives in the spec, so it will just be easier to move to that then.

src/definitions/directive.ts
```diff
@@ -0,0 +1,211 @@
...
+  ) {}
+}
+
+withNexusSymbol(NexusDirectiveUse, NexusTypes.DirectiveUse)
+
+/** Defines a directive, which can be used positionally when generating the SDL */
+export function directive<DirectiveName extends string>(
+  config: NexusDirectiveConfig<DirectiveName>
+): NexusDirectiveDef<DirectiveName> {
+  assertValidName(config.name)
```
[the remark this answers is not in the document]
tgriesser: Actually recently found that we don't do this for field names, but we should 😬

tgriesser:
> Is there an estimated time for this new feature? 👀

Going to probably work on it a bit today/tomorrow.

davidsoderberg:
Any progress on this one?

yasaichi:
What's a blocker of merging this PR?

victoriris:
@davidsoderberg & @yasaichi just to let you know... I needed this for Apollo Federation support and ended up switching to type-graphql a month ago

davidsoderberg:
I also ended up switching to type-graphql... :(

src/builder.ts
redbar0n: @tgriesser what would be the shortest path to unblock this pull request so it can go through?

zingerj:
Hey all, what's the latest on getting this merged? Looks like there's one failing test; is that it to get this across the line? Thanks for your work here!

src/builder.ts
redbar0n: @zingerj this is the blocker.

YElyousfi:
When can we expect this to get merged? Federation relies on directives.

meabed:
This is great feature - thank you!! 
Would it be ready to merge soon?

yasaichi:
To those who can't wait this PR merged
You have the following alternative libraries supporting directives:

- [TypeGraphQL](https://typegraphql.com/docs/directives.html) (mentioned above)
- [Pothos](https://pothos-graphql.dev/docs/plugins/directives) (formerly GiraphQL)

BahaaZidan:
Please merge this!

src/builder.ts
tgriesser: Going to punt on this for now - going to assume that if you need to modify the ASTNode you can use something else more specialized like graphql-tools

tgriesser:
This is merged, and is now released under `nexus@1.4.0-next.7`, no docs yet other than in the [PR description](https://github.com/graphql-nexus/nexus/pull/952#issue-949724569). 

No official plugin for federation, but the primitives are there - feel free to open a PR if you make progress on something generic and would like to add a plugin to the core.

Additionally, please give any feedback or let me know if you run into issues with it, will get this into an official `1.4.0` soon

meabed:
@tgriesser - Thank you! any date we expecting this to be released?

cqhung1412:
@tgriesser - Please release this soon with the docs.

unrevised6419:
Hey folks! Thanks for working on this feature for Nexus!

Should there be any type of completion for the directive args when using it? 🤔 

image

image

I've augmented it using this

```ts
declare module 'nexus/dist/core' {
  export interface NexusDirectiveDef {
    (arg: CacheHint): NexusDirectiveUse<"cacheControl">;
  }
}
```

PS: Does anyone know when a new minor release will be done with this feature?

Not related: I'm using this feature to implement cache control in Strapi. I started with using resolvers, but I think the directive approach is much better.

https://github.com/strapi/strapi/pull/20254

Issue #269, opened 2019-10-15
Multiple schemas typing issues


alizhdanov:
Hello,
I have two schemas in one project (just for example - `schema1` and `schema2`). Each of them has it's own `nexus-typegen` generated file. Types for `schema1` are working fine, the issue with `schema2`. For instance when I wanna create new `mutationField`:

```
mutationField('createDoughnut', {
  type: 'Doughnut',
});
```

It'd show typescript error that `type '"Doughnut"' is not assignable to type ...`. The issue that this type defined in `schema2`, but not in `schema1`. But types always checked from `schema1`.

The issue seems to be, because of `nexus` is dependant on global types defined in `@types/nexus-typegen__nexus-core`. 
```
declare global {
  interface NexusGen extends NexusGenTypes {}
}
```

In my case I generate, which mean that only one of the file will be applied:
* `@types/nexus-typegen__nexus-core/schema1.ts
* `@types/nexus-typegen__nexus-core/schema2.ts

Is it possible somehow to link types manually?

valentinpalkovic:
Therefore it is currently not usable with graphql-modules, if you have at least two nexus modules

tgriesser:
The assumption of nexus is that there is one schema per typescript project. If you use a lerna/yarn workspaces style monorepo, it should work if you have the generated types emitted into separate projects. I currently use this approach with an "admin" schema which is a superset of the "public" schema. In order for this to work you need to manually specify the output paths for these modules.

I'm not sure about graphql-modules.

valentinpalkovic:
@tgriesser Why is typescripts declaration merging not working? I do not get it exactly.

tgriesser:
Because the shape of how we define types currently makes them un-mergeable. IIRC the original thought was to centralize on a single interface and then pass that down as a default parameter to be able to get to the types via conditionals.

I'm not certain if this is actually necessary still with how we're inferring them now via the globals and "materializing" the types dynamically via `infer`, it's been on the todo list of design assumptions to revisit.

PR #967, opened 2021-08-17
feat: ConfiguredTypegen for splitting global types & configuring input type emission

tgriesser:
Adds a new signature to `outputs.typegen` from `string | null` -> `string | null | ConfiguredTypegen`, where `ConfiguredTypegen` is defined as:

```ts
export interface ConfiguredTypegen {
  /** 
   * Path for the generated type defs 
   */
  outputPath: string
  /**
   * Determine the path the "globals" are output, useful when you have a monorepo setup and need to isolate
   * the globals from the rest of the types in order to have multiple schemas/ts projects
   */
  globalsPath?: string
  /**
   * If globalsPath is defined, these headers are added to the "globals" generated file, rather than the
   * typegen generated file
   */
  globalsHeaders?: string[]
  /**
   * If "true", declares dedicated interfaces for any inputs / args
   *
   * @default true
   */
  declareInputs?: boolean
}
```

The `globalsPath` allows us to separate the location of the emitted files from a single file into two files, one with the 
`declare global {` wrapping, and the other with the normal generated types. This will help with issues like #269, where you would like multiple schemas in the same codebase - by separating the types you might want to use across multiple projects from the ones that are specific to the local schema, it becomes possible to address that sort of case.

Additionally, adds an option `declareInputs` which will emit dedicated types for use on the server for the inputs. We will default to `true` moving forward since this behavior is beneficial to end users, and allow them to opt out (doesn't need to be a breaking change).


tgriesser:
Merging for now to kick the tires on it a bit, can clean up / make any modifications before the next release

PR #1041, opened 2022-02-03
feat: Use ReadonlyArray in typings

vinassefranche:
fixes https://github.com/graphql-nexus/nexus/issues/944

Concerning the outputs, using `ReadonlyArray` instead of `[]` or `Array` should not break anything so I made the change without adding any config.
Concerning the inputs, it can break the build of existing projects so I added the `useReadonlyArrayForInputs` config for `makeSchema`.

Tell me if:
- I added the config in a terrible way. I did what I supposed was fine but I might completely wrong
- You think I should have a global config for the inputs and the outputs
- You think I should add tests for the config I added
- Anything else is wrong with my P.R. :-)


tgriesser:
Definitely agree on supporting Readonly Array on output return types. I had done a little work on this at one point and had originally something like:

```
type Arr<T> = Array<T> | ReadonlyArray<T>
```

so that the output was a little less verbose, but I guess as you point out there isn't a reason we'd need the `Array<T> | ` for output types, so I think it's probably fine to keep as you have it unless I'm forgetting something.

Regarding the input types, agree it should be opt-in, so as to not require a major version bump. I had started to consolidate options for managing the typegen into the [`ConfiguredTypegen` option](https://github.com/graphql-nexus/nexus/blob/a003ff074fdc1d86abaa6fad87f7376823c37418/src/builder.ts#L179-L198), as a place to consolidate options like this that will affect the typegen in the future:

```ts
makeSchema({
  outputs: {
    outputPath: './path/to/nexus.gen.ts',
    useReadonlyArrayForInputs: true
  }
})
```

Mind seeing if you can add the option there / adding a test and we can get this merged.

PR #1042, opened 2022-02-04
feat: Run both formatTypegen and prettier formatter if given

vinassefranche:
Hi!
It's not stated clearly in the doc that giving `formatTypegen` in the config prevents the `prettierConfig` from being used and I think it's fine to have both used the code.
Because of how the code is currently, I have to import `typegenFormatPrettier` with a nested import (`import { typegenFormatPrettier } from 'nexus/dist/typegenFormatPrettier';`) and use it in my `formatTypegen` function.

Tell me if you think it's ok to change this behavior.


tgriesser:
@vinassefranche Seems like a reasonable change, mind adding a test for it?

vinassefranche:
@tgriesser thanks for your review!
Instead of adding a test for it, I updated how both prettier and some import updates were done in some existing test to use both `prettierConfig` and `formatTypegen`. I think it enough to test this modification. Do you agree?
I had to be more precise to find only the imports by searching for `from "nexus"` instead of just `"nexus"` as there was another `"nexus"` occurrence in the schema, FYI.

tgriesser:
@vinassefranche looks like this conflicted when the other PR you had that I merged, mind updating when you have a minute?

vinassefranche:
@tgriesser this should be good now!

PR #1057, opened 2022-03-10
docs: Update comment for `shouldGenerateArtifacts`

zingerj:
Update comment to reflect the current behavior as per [`src/typegenUtils.ts`](https://github.com/graphql-nexus/nexus/blob/bc12ca0f8e704de0b7e911fa6352e8f2bb48cb22/src/typegenUtils.ts#L11). Ideally this behavior would be updated to reflect the original comment as most use cases won't need artifacts generated outside of `development`, but that would introduce a breaking change so probably needs to wait for a major release.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -278,7 +278,7 @@ export interface BuilderConfigInput {
       }
   /**
    * Whether the schema & types are generated when the server starts. Default is !process.env.NODE_ENV ||
-   * process.env.NODE_ENV === "development"
+   * process.env.NODE_ENV !== "production"
    */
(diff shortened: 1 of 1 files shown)


tgriesser:
Thanks @zingerj - I think that's how I had it originally, it changed here https://github.com/graphql-nexus/nexus/pull/884 and I didn't really think about it hard enough I guess. 

I think it should probably work as it was before that change, since you don't want it generating when you have something like `NODE_ENV=test`, but let's at least update this for now and maybe we can get it right in 2.0

PR #1072, opened 2022-04-08
feat: allow after build hook to modify the schema

hemache:
e.g. graphql middlewares


tgriesser:
Interesting change, seems reasonable, mind adding a test to cover it?

tgriesser:
Actually, I think we want to add a new hook here, `onBeforeSchemaBuild` which takes the config that will be passed to `new GraphQLSchema`, and can modify that, which is better than needing to build a whole new schema.

Issue #1069, opened 2022-04-05
Option To Omit Connection Descriptions In Connection Plugin


danielschwartz:
We're using Nexus in our production application (love it btw) and we noticed that when using the connection plugin a bunch of the types (e.g. `PageInfo`, `<Object>Edge`, etc.) have hardcoded links to `facebook.github.io`.

Would it be possible to add an option to the plugin config to omit these links from the generated schema? We'd prefer not to have any links to Facebook in there, even if it is to a github page.

I can create a PR if it's helpful, I just don't have the development environment setup locally, so figured it might be easier for a contributor to do this real quick.

tgriesser:
Would it be ok if the URL was changed to https://relay.dev/graphql/connections.htm?

PR #1083, opened 2022-04-28
chore: change facebook.github.io to relay.dev links

danielschwartz:
As I brought up in #1069 we'd prefer not to have the word `facebook` in our publicly available schema. @tgriesser asked if changing the links to `relay.dev` was acceptable and we agreed it was. This PR simply changes any `facebook.github.io` links to their respective `relay.dev` links. `facebook.github.io/relay` already redirects to `relay.dev` so this should essentially be a no op from a documentation perspective.


tgriesser:
Thanks @danielschwartz!

PR #1128, opened 2022-08-22
chore: remove irrelevant paragraph in docs

alexkunin:
Resolver shorthands were removed in v0.18.0, docs need to be updated.

See https://github.com/graphql-nexus/nexus/blob/main/src/messages.ts#L31


tgriesser:
Thanks @alexkunin!

PR #1141, opened 2022-10-28
fix: add missing subcribe when rebuildOutput, solve subcription megreSchema problem

GavinXue:
There is a "Subscription field must return Async Iterable. Received: undefined" problem when merge Subcription type.  It can reproduce here: https://stackblitz.com/edit/graphql-xtfnno?file=index.mjs

I found that the mergeSchema's step `rebuildOutputDefinition` doesn't return subscribe, so it the subscribe will replaced by undefined. this commit add subscribe to return.

Hope it helpful.


JWambaugh:
Can we get this merged? I've been slamming my head against this bug all day.

PR #1161, opened 2023-05-25
fix: Remove deprecated graphql method assertValidName and replaced with assertName

toinhao1:
Remove deprecated graphql method assertValidName and replaced with assertName


toinhao1:
@tgriesser anyway we could get this merged and deployed?

